PR #2, opened 2015-06-30
handle side-effects in Try Monad

gkhom:


src/main/scala/nl/knaw/dans/api/sword2/Test.scala
```diff
@@ -0,0 +1,19 @@
+package nl.knaw.dans.api.sword2
```
janvanmansum: I think this one's not intended for the code base, right? :)

gkhom: @janvanmansum Indeed, I've removed it.

pom.xml
```diff
@@ -52,6 +52,11 @@
       <artifactId>scala-library</artifactId>
       <version>2.11.7</version>
     </dependency>
+    <dependency>
+      <groupId>net.lingala.zip4j</groupId>
```
janvanmansum, line 56: A customer of the SWORD 1.3 API reported a problem that turned out to be related to his use of the lingala zip4j library. Here is a discussion topic about the problem that I found after doing a web-search:
http://www.lingala.net/zip4j/forum/index.php?topic=268.0
I'm not sure if zip4j is in the wrong, just mentioning it, so you know it may cause problems.

src/main/scala/nl/knaw/dans/api/sword2/Authentication.scala
```diff
@@ -25,7 +25,7 @@ object Authentication {
       throw new SwordError("http://purl.org/net/sword/error/MediationNotAllowed")
     }
     // temporary short-circuit
-    if (!(auth.getUsername == SwordProps.get("user")) || !(auth.getPassword == SwordProps.get("password"))) {
+    if (!(auth.getUsername == SwordProps("user")) || !(auth.getPassword == SwordProps("password"))) {
```
janvanmansum, line 28: A rather complicated boolean expression. How about

```
 auth.getUsername != SwordProps("user") || ...
```

?

Why a _temporary_ short-circuit? Are you referring to the upcoming implementation using LDAP?

src/main/scala/nl/knaw/dans/api/sword2/ServiceDocumentManagerImpl.scala
```diff
@@ -30,7 +30,7 @@ class ServiceDocumentManagerImpl extends ServiceDocumentManager {
     val sc: SwordCollection = new SwordCollection
     sc.setTitle("DANS EASY Data Collection")
     sc.addAcceptPackaging("http://easy.dans.knaw.nl/schemas/index.xml")
```
janvanmansum, line 32: What would be a better URL to use as ID for our DANS/bagIt packaging format?

src/main/scala/nl/knaw/dans/api/sword2/SwordID.scala
```diff
@@ -17,30 +17,33 @@ package nl.knaw.dans.api.sword2
 
 import org.swordapp.server.SwordError
 
+import scala.util.{Failure, Success, Try}
+
 object SwordID {
-  def generate: String = {
-    val id: Long = System.currentTimeMillis
+  def generate: Try[String] = synchronized {
```
janvanmansum, line 23: _synchronized_ could become a bottle neck here, especially if we sleep for 2 ms in it. It will do for now, though.

src/main/scala/nl/knaw/dans/api/sword2/Authentication.scala
gkhom, line 28: Yes, this is temporary since it only works for one user (configured in the properties). The LDAP implementation will make this obsolete.

PR #13, opened 2015-09-01
Easy deposit git

janvanmansum:
- Made git-support configurable
- Deleted tests (should be replaced by unit tests in the future)
- Added logging
- Made log level configurable
- Made EASY_DEPOSIT_HOME a servlet init parameter so that the configuration can be done
  through a Tomcat context fragment. It is now possible to run two instances on the same
  server, which comes in handy when testing both the deposit and the darkarchive service
  on deasy


janvanmansum:
@GeorgiKhomeriki @PaulBoon @jo-pol @ekoi please review.

gkhom:
This PR LGTM :+1:

PR #14, opened 2015-09-02
Easy deposit improved move

janvanmansum:
The move was implemented as a copy. File.renameTo does an actual "move" operation.

@GeorgiKhomeriki @PaulBoon @ekoi @jo-pol please review.


janvanmansum:
Please, first review EASY_DEPOSIT_GIT, as this pull request is based on it.

gkhom:
I didn't know you could use `renameTo` like that, looks good :+1:

janvanmansum:
I didn't know before today, either. Just googled it, tested it on deasy and it seems to work.

PR #22, opened 2015-09-09
Fixed example

janvanmansum:
Fixed example so that it can be used to test the complete ingest flow.

@GeorgiKhomeriki


gkhom:
:+1:

PR #28, opened 2015-09-18
Fix for Swordapp library

janvanmansum:
- POST-ing to the SE-IRI is not supported by the official library, so we forked it to add support
- Refactored the code so that continued deposit is now routed through the container manager implementation
- Fixed the statement implementation

@GeorgiKhomeriki , please have a look.


src/main/scala/nl/knaw/dans/api/sword2/StatementManagerImpl.scala
```diff
@@ -31,22 +29,27 @@ class StatementManagerImpl extends StatementManager {
...
     val dir = new File(SwordProps("data-dir"), id)
     if (!dir.exists) throw new SwordError(404)
     val git = Git.open(dir)
-    val head = git.getRepository.resolve(Constants.HEAD)
-    val headTag = git.tagList().call().find(_.getObjectId.equals(head))
-    if (!git.status().call().isClean || headTag.isEmpty)
-      "state=DRAFT"
-    else
-      headTag.get.getName
+    val tag = git.describe.call()
```
janvanmansum: Doesn't have the correct result. See #29

PR #35, opened 2015-10-07
Absense of the swordDepositAllowed attribute in the user entry led to a NullPointerException instead of the correct 403 Forbidden

janvanmansum:


janvanmansum:
@GeorgiKhomeriki @jo-pol @PaulBoon @ekoi please have a look.

gkhom:
Looks like a simple fix :+1:

Issue #38, opened 2015-10-24
Store password as hash instead of as clear text in one-user set-up


janvanmansum:
@DANS-KNAW/easy 
EASY-921

Storing passwords in clear text is an anti-pattern. In the single-user set-up it is still done this way. Instead of the password itself store a hash. Some research in how to this properly may be in order, e.g. https://crackstation.net/hashing-security.htm

jo-pol:
Can this one be closed now that #50 is merged?

janvanmansum:
Implemented in #50

PR #50, opened 2016-02-09
Hashed password, issue #38

jo-pol:
@DANS-KNAW/easy 
The code of the first commit in this pull request is partly advised, partly commented as anti-pattern in the [documentation](https://crackstation.net/hashing-security.htm) recommended in the issue. As we have just one internal account for an internal service the implementation seems good enough.

However, I could not figure out a corresponding command line, so the final implementation performs another calculation, the corresponding command line is documented in the unit test.

### DIFF
--- a/src/test/scala/nl/knaw/dans/api/sword2/AuthenticationSpec.scala
+++ b/src/test/scala/nl/knaw/dans/api/sword2/AuthenticationSpec.scala
@@ -1,19 +1,18 @@
-/*******************************************************************************
-  * Copyright 2015 DANS - Data Archiving and Networked Services
-  *
-  * Licensed under the Apache License, Version 2.0 (the "License");
-  * you may not use this file except in compliance with the License.
-  * You may obtain a copy of the License at
-  *
-  *   http://www.apache.org/licenses/LICENSE-2.0
-  *
-  * Unless required by applicable law or agreed to in writing, software
-  * distributed under the License is distributed on an "AS IS" BASIS,
-  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-  * See the License for the specific language governing permissions and
-  * limitations under the License.
-  ******************************************************************************/
-
+/**
+ * Copyright (C) 2015-2016 DANS - Data Archiving and Networked Services (info@dans.knaw.nl)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package nl.knaw.dans.api.sword2
 
 import nl.knaw.dans.api.sword2.Authentication.hash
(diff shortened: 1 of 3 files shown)


src/test/scala/nl/knaw/dans/api/sword2/AuthenticationSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package nl.knaw.dans.api.sword2
+
+import nl.knaw.dans.api.sword2.Authentication.hash
+import org.scalatest.{FlatSpec, Matchers}
+
+class AuthenticationSpec extends FlatSpec with Matchers {
+
```
janvanmansum, line 22: Readable and concise unit test.

src/test/scala/nl/knaw/dans/api/sword2/AuthenticationSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+  val command = "echo -n 'SomePassword' | openssl sha1 -hmac 'someUserNameAsSalt' -binary | base64"
+  val output = "WjYViDQOdGR8V1kkTs900ZfoLXU="
+
+  "hash" should s"return same as: $command" in {
```
janvanmansum, line 26: The output of the test run will be less readable. It will print the command for every test that is run. More informative would be something like

```
  "hash" should "return stored hash if password is correct" in { // etc ...
  "hash" should "not return stored hash if password is incorrect" in { // etc ...
```

janvanmansum:
Not sure what you mean by "the corresponding  command line". A salted password would have been nice, but it is not a high priority. We'll use a generated, secure password in the production environment anyway.

src/test/scala/nl/knaw/dans/api/sword2/AuthenticationSpec.scala
jo-pol, line 26: As there are many algorithms to compute some hash and convert the result to string (hex, base64), I wanted to get documented what was applied. Same thoughts behind "corresponding command line". You need to know what to configure in the application.properties, perhaps assumed knowledge by you but it was a mystery for me.

janvanmansum, line 26: Sure, documenting that is a great idea. If you read my comment carefully you'll see it it is about printing that command line with every test. The test cases are well chosen, but the description might have been better. Not a big thing, just nitpicking, as usual :)

jo-pol, line 26: hrm, just tried 

```
import sys.process._
val storedPassword =
  "echo -n 'SomePassword'" #|
  "openssl sha1 -hmac 'someUserNameAsSalt' -binary" #|
  "base64" !!
```

but the echo adds an `\n` which it should not because of the `-n`

janvanmansum, line 26: Strange, but I don't think we should build platform dependencies into the unit tests, anyway.

jo-pol, line 26: @DANS-KNAW/easy 

To be able to configure the password in `application.properties` there should be some command line equivalent that computes the same hash as is done in the code. So imho it is reasonable to have the command sequence in the test. If someone is tempted to install the suite of EASY components on windows we might run into more problems than just this command sequence. 

The actual problem were the single quotes around the command line arguments that are interpreted by a shell but not by scala's ProcessBuilder.

janvanmansum, line 26: - It is not just about using Linux, Mac or Windows, it's also about making assumptions about which exact commands are installed.
- I should hope that our depedencies are limited to the Java platform, but if you think there are more OS-dependencies I would like to have those pointed out to me.

PR #52, opened 2016-04-22
EASY-1042 Lege Slug-header zorgt voor lege slug

janvanmansum:
A blank slug is now treated as no slug at all.


janvanmansum:
@lindareijnhoudt @jo-pol

PR #57, opened 2016-07-05
Rename deposit to sword2

rvanheest:
~~fixes EASY-~~
#### When applied it will
- rename the `Easy-Deposit` project to `Easy-Sword2`
#### Where should the reviewer @DANS-KNAW/easy start?
- Please do not only focus on the changes I made, but also to the rest of the code/text/...! Did I miss any occurrences of `deposit` that need to be changed to `sword2`? Not that this is not trivial, as `deposit` was used kind of ambiguously in this project: it can refer to a 'real' deposit or refer to the project's name. Only the latter should be changed to `sword2`!
- **IF YOU HAVE A FORK OF THIS PROJECT, I KINDLY REQUEST YOU CHANGE THE NAME OF THIS FORK ACCORDINGLY!**
- **PLEASE CHANGE THE REMOTE(S) ON YOUR LOCAL CLONES TO POINT TO THE NEW NAME/URL.** [See here for instructions.](https://help.github.com/articles/changing-a-remote-s-url/)
- **PLEASE CHANGE THE NAME OF THE FOLDER OF YOUR LOCAL CLONE ACCORDINGLY.**
#### How should this be manually tested?

This should be tested together with other PRs yet to come in Easy-DTAP and Easy-Build. The other related PR's just changes some textual references and do not influence the code. To test this:
- check out these PRs
- do a `mvn clean install`
- `vagrant destroy && vagrant up` should (after lots of waiting) create a new instance of `deasy`
- _optionally:_ you can try the procedures as described in `README.md` to ensure that everything is still working according to plan.
#### related pull requests on github

| repo | PR | note |
| --- | --- | --- |
| easy-build | [PR#7](https://github.com/DANS-KNAW/easy-build/pull/7) | small change in `pom.xml` |
| easy-dtap | [PR#19](https://github.com/DANS-KNAW/easy-dtap/pull/19) |  |
| easy-license-creator | [PR#6](https://github.com/DANS-KNAW/easy-license-creator/pull/6) | small change in `Functional requirements.md` |
| easy-stage-dataset | [PR#87](https://github.com/DANS-KNAW/easy-stage-dataset/pull/87) | small change in `README.md` |
| easy-ingest-flow | [PR#47](https://github.com/DANS-KNAW/easy-ingest-flow/pull/47) | small changes in `pom.xml` and `README.md` |
[...]

PR #56, opened 2016-06-23
Bugs found in easy-deposit

rvanheest:
@janvanmansum small change...

PR #59, opened 2016-07-06
Fixed two bugs

janvanmansum:
#### When applied it will
- Fix the regression bug introduced by 50c7b5db9deb6306e9bbf4448a6521f194fad690. The problem was that  `deposit.properties` initially does not exist, which causes the deposit to fail when the code tries to load it.
- Fix a nested Try construct which caused above error not to be reported back to the client. (A `Failure` object was wrapped inside a `Success`.) 

`DepositProperties.scala`

In `easy-dtap`:

```
./deploy-role easy_sword2
```

In `easy-sword2/src/test/resources`:

```
./send-simple.sh simple/example-bag.zip http://deasy.dans.knaw.nl/sword2/collection/1
```

If you want to check that an exception in `DepositProperties.set` does show up in the response sent to the client, you could rebuild the project with a hardcode `throw new RuntimeException("foo")` in that method and perform the test again.

none

### DIFF
--- a/src/main/scala/nl/knaw/dans/api/sword2/CollectionDepositManagerImpl.scala
+++ b/src/main/scala/nl/knaw/dans/api/sword2/CollectionDepositManagerImpl.scala
@@ -43,12 +43,12 @@ class CollectionDepositManagerImpl extends CollectionDepositManager {
     if(iri != SwordProps("collection.iri")) throw new SwordError("http://purl.org/net/sword/error/MethodNotAllowed", 405, s"Not a valid collection IRI: $iri")
   }
 
-  private def setDepositStateToDraft(id: String, userId: String): Try[Unit] = Try {
+  private def setDepositStateToDraft(id: String, userId: String): Try[Unit] =
     DepositProperties.set(
       id = id,
       stateLabel = "DRAFT",
       stateDescription = "Deposit is open for additional data",
       userId = Some(userId),
       lookInTempFirst = true)
-  }
+
 }
--- a/src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
+++ b/src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
@@ -65,8 +65,8 @@ object DepositProperties {
   private def readProperties(f: File) = {
     val ps = new PropertiesConfiguration()
     ps.setDelimiterParsingDisabled(true)
-    ps.load(f)
-
+    if(f.exists) ps.load(f)
+    ps.setFile(f)
     ps
   }
 }


src/main/scala/nl/knaw/dans/api/sword2/CollectionDepositManagerImpl.scala
```diff
@@ -43,12 +43,12 @@ class CollectionDepositManagerImpl extends CollectionDepositManager {
     if(iri != SwordProps("collection.iri")) throw new SwordError("http://purl.org/net/sword/error/MethodNotAllowed", 405, s"Not a valid collection IRI: $iri")
   }
 
-  private def setDepositStateToDraft(id: String, userId: String): Try[Unit] = Try {
+  private def setDepositStateToDraft(id: String, userId: String): Try[Unit] =
```
janvanmansum, line 46: Called method already returns a `Try[Unit]`, wrapping it again in a `Try` will make `Failure`s invisible to the client.

src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
```diff
@@ -65,8 +65,8 @@ object DepositProperties {
   private def readProperties(f: File) = {
     val ps = new PropertiesConfiguration()
     ps.setDelimiterParsingDisabled(true)
-    ps.load(f)
-
+    if(f.exists) ps.load(f)
```
janvanmansum, line 68: Only load the file if it exists.

src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
```diff
@@ -65,8 +65,8 @@ object DepositProperties {
...
+    ps.setFile(f)
```
janvanmansum, line 69: We need to set the file name to be able to save it initially.

rvanheest:
Hey, guess what, that's what I mean with "_I think I fixed the bug with get.sh in commit 50c7b5d,_ **_but I can't test this until #57 is merged**_. _On hold again._" (see [this comment](https://github.com/DANS-KNAW/easy-sword2/pull/56#issuecomment-230441492)). If it ain't tested, don't merge it :smile:

Also, there are some unsolved points mentioned in #56 ([see here](https://github.com/DANS-KNAW/easy-sword2/pull/56#issuecomment-228048107)). If you did those while finding this bug, please mark them as solved as well.

src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
rvanheest, line 68: Quick question: are you sure we still want an empty `PropertiesConfiguration` when the file doesn't exist? Haven't looked at the context, just wondering!

janvanmansum:
OK, fair enough.

src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
janvanmansum, line 68: This methods sets the current deposit state, so if `deposit.properties` doesn't exist it must be created.

rvanheest, line 68: Looked in the `PropertiesConfiguration` source code and this is exactly what we want. Great job! Sadly we need to do this ourselves due to the `setDelimiterParsingDisabled(true)` thingie...

rvanheest:
BTW, I see now that `stackTraceToString` isn't used in this file. Maybe you can delete this in this PR as well?

PR #61, opened 2016-08-04
EASY-1094: Files in the fetch.txt are resolved before the bag validity is checked

vesaakerman:
fixes EASY-1094
#### When applied it will

Resolve the files in fetch.txt (reads the files into the temporary bag directory). Now the bag contents can be validated.  When the files are not yet resolved, the bag is not valid.

In this package there are instructions about how to test using command line scripts. 

Note: to check the validity of the urls in the fetch.txt a regular expression is stored in application.properties.  Now we have a very permissive regular expression there, but that will be changed when we have decided on  what are acceptable urls. 

@janvanmansum 
- [x] Process code review comments
- [x] Unit tests


pom.xml
```diff
@@ -108,6 +108,11 @@
             <groupId>org.scalatest</groupId>
             <artifactId>scalatest_2.11</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.jsuereth</groupId>
+            <artifactId>scala-arm_2.11</artifactId>
+            <version>2.0.0-M1</version>
```
janvanmansum: `version` not necessary, is inherited from parent pom. By the way, please upgrade parent pom to version `1.31`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -64,14 +69,15 @@ object DepositHandler {
...
+      _        <- resolveFetchItems(bagitDir)
       _        <- checkBagValidity(bagitDir)
       _        <- DepositProperties.set(id, "SUBMITTED", "Deposit is valid and ready for post-submission processing", lookInTempFirst = true)
       _        <- commitSubmitted(git, tempDir)
       dataDir  <- moveBagToStorage()
     } yield ())
     .recover {
       case InvalidDepositException(_, msg, cause) =>
-        log.warn(s"[$id] Invalid deposit", cause)
+        log.error(s"[$id] Invalid deposit", cause)
```
janvanmansum, line 80: OK, could also be `error`. I am not sure about the best criterion here so, fine, change it.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -160,6 +166,38 @@ object DepositHandler {
     }
   }
 
+  private def resolveFetchItems(bagitDir: File)(implicit id: String): Try[Unit] = Try {
+    log.debug(s"[$id] Checking fetch.txt")
```
janvanmansum: When I encountered this line in the logs, I wasn't really sure if that meant a `fetch.txt` was found. Apparently, it does not mean that, necessarily. I think the message should be clear about that.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -160,6 +166,38 @@ object DepositHandler {
...
+    getFetchTxt(bagitDir).foreach(fetchText =>
+      { log.debug(s"[$id] Resolving files in fetch.txt")
+        fetchText.asScala.foreach(item =>
+        getUrl(item.getUrl).foreach(url =>
+          Using.urlInputStream(url).foreach(src =>
+            Files.copy(src, Paths.get(bagitDir.getAbsolutePath, item.getFilename)))))
+      }
+    )
+  }
+
```
janvanmansum, line 177: What type does `getFetchTxt` return? What does `foreach` mean on that type? (`foreach` suggests that collection type is used, which is not actually the case here. This is rather confusing.)

What happens if a url is syntactically incorrect? What happens when it is correct but cannot be dereferenced? Does it result in a clear error message, both to the client and in the log?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -160,6 +166,38 @@ object DepositHandler {
...
+  private def getFetchTxt(bagitDir: File)(implicit id: String): Option[FetchTxt] = Option {
+    getBagFromDir(bagitDir).getFetchTxt
+  }
+
+  private def getBagFromDir(dir: File)(implicit id: String): Bag = {
```
janvanmansum: `bagFactory.createBag` is now called in two places in this file. This calls for a refactoring!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -160,6 +166,38 @@ object DepositHandler {
...
+      bagFactory.createBag(dir, BagFactory.Version.V0_97, BagFactory.LoadOption.BY_MANIFESTS)
+    } catch {
+      case e: Throwable => throw new FailedDepositException(id, "Failed to create a bag when resolving Fetch Items", e)
+    }
+  }
+
+  private def getUrl(url: String)(implicit id: String): Try[URL] =  {
+    val urlPattern = SwordProps("fetch.allowed-url-pattern").r
+    urlPattern findFirstIn url match {
+      case Some(_) => Success(new URL(url))
```
janvanmansum: `findFirstIn` suggests that you are looking of a URL in some larger text. I think we actually want to check that the `url` argument **is** a valid URL.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
janvanmansum: Come to think of it, I think the log line is okay, but you should actually **do** what it says: first check the fetch.txt and only try the downloads if all the lines are okay. So: fail fast, if part of the lines are okay, still reject the whole bag.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
janvanmansum: @vesaakerman Added (your commit message above) a check list for the items of work that need to be done before this can be merged. 

@DANS-KNAW/easy for further review.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: My intention was not to check the contents of the fetch.txt before downloading the files.  I added "Checking fetch.txt" to make it clear that the program is checking whether there are files to be fetched.  Only in the case that there is a fetch.txt file, shall the "Resolving files in fetch.txt" text be shown.  I wonder if it is needed to check the contents of the fetch.txt file before downloading the files.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman, line 177: _getFetchTxt_ returns type Option[FetchTxt]. In this case _foreach_ gives the _FetchTxt_ if it is present and valid.  Scala prefers _foreach_ here over _map_.  I have understood that this is a common way in Scala (to handle either with _map_ or with  _foreach_ an Option type.

If the url is syntactically incorrect or when it can't be dereferenced, the client gets an error message "Invalid url in Fetch Items ($url)".  In the log the exact cause of the error is shown.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: I think that we have to call _bagFactory.createBag_ twice, because the contents of the bag directory has been changed after fetching the files in the fetch.txt.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: In Scala Regex we have findFirstIn and findAllIn.  I agree that the name of the method doesn't tell exactly what we are actually doing here: checking if the the given url matches the pattern against which it is compared. I still would like to to use this method, okay?

janvanmansum: The name of the method tells exactly what it does. It just isn't what we really want to do. I think the following is better:

```
if (urlPattern.pattern.matcher(url).matches) // Success(new URI(url)) 
else // Failure(...)
```

Although it is a bit more verbose, it expresses and does what we need: match the complete URL-string against the configured pattern. If we choose the wrong `urlPattern` we would get unexpected results with the code as it is now. Compare:

```
 "http://.*".r.pattern.matcher("not a valid url http://www.google.com").matches() // return false
 "http://.*".r.findFirstIn("not a valid url http://www.google.com").isDefined // returns true
```

Since we didn't put in the starting "^" and closing "$" we get an incorrect result in the second line.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
janvanmansum: - So you intended to say: "Checking if bag contains fetch.txt" or something. 
- Failing fast has the advantage that it becomes much easier to find the cause of an error.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
janvanmansum, line 177: Okay, I didn't know that `foreach` was also idiomatic on `Option`, even preferred, but at face value both `foreach` and `map` are kind of counter-intuitive. I guess this just takes some getting used to then.

It would be better to clearly distinguish between a syntactic error and a dereferencing error. Could you do that?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
janvanmansum: - That may be correct; something to check.
- What I actually meant was that the call to `bagFactory.createBag` is now in two places in the code in de function `getBagFromDir` and in `checkBagValidity`. In that second function you might call `getBagFromDir`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Well yes, I was assuming that we would make sure that there is a valid pattern in the properties file.  But yes, this is safer, and also more intuitive.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman, line 177: Handling `Options` and `Try's` with `map` and `foreach` is truly compact and neat. But I agree that it is not really intuitive when you have only one member or zero.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -160,14 +166,61 @@ object DepositHandler {
...
+    log.debug(s"[$id] Checking validity of urls in fetch.txt")
+    fetchText.asScala.foreach(item =>
+      checkUrlValidity(item.getUrl))
+  }
+
+  private def checkUrlValidity(url: String)(implicit id: String): Unit = {
+    // check if the url is syntactically correct
+    try {
+      new URL(url)
+    } catch {
```
janvanmansum, line 192: You shouldn't catch **all** `Throwable`s here. What if `url` were to be `null`, is that a syntax error? What if for some reason the `URL` class were not to be found so we got a `NoClassDefFoundError`, is that a syntax error? As a general rule try to be specific as possible and practicable when catching exceptions.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -160,14 +166,61 @@ object DepositHandler {
...
     val validationResult: SimpleResult = bag.verifyValid
     if (validationResult.isSuccess) Success(Unit)
     else Failure(new InvalidDepositException(id, validationResult.messagesToString))
   }
 
+  private def getBagFromDir(dir: File)(implicit id: String): Bag = {
+    try {
+      bagFactory.createBag(dir, BagFactory.Version.V0_97, BagFactory.LoadOption.BY_MANIFESTS)
+    } catch {
+      case e: Throwable => throw new FailedDepositException(id, "Failed to create a bag", e)
```
janvanmansum: Same problem as above: catching all `Throwable`s not a good idea. (I thought of one more example: what about `OutOfMemoryException`, okay, now I'll stop :)

rvanheest: Agree, but why not use `scala.util.Try` instead?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman, line 192: You are right. I have now `MalformedURLException`, which catches syntactically wrong urls, including `null`.  All the other exceptions will be thrown higher up as `Unexpected failure in deposit` with the original cause.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: I agree, too.  I removed the whole catch, and thus now the exception will be thrown higher up as `Unexpected failure in deposit` with the original cause.

vesaakerman: I actually had first `scala.util.Try`.  But there is no advantage of that here, because these exceptions cause the whole process to stop.

PR #64, opened 2016-10-10
Implementation of "virtually valid" bags.

vesaakerman:
fixes EASY-1096
#### When applied it will
- download those files in fetch.txt that refer outside the bag store
- validate the bag
- if only 'missing payload files' validation errors, proceed
- if missing payload files are the same as the files in fetch.txt referring to the bag store, proceed
- verify the checksums of these files against the checksums in the referred bag

See testing instructions in the project description. Notice: because at the moment there is no feedback about the uri of the newly created bag, you have to use the test bags in src/test/resources/simple-sequence.  Now you also have to copy the contents of the src/test/resources/bag-store directory to the /data/bag-store directory of the target server.

| repo | PR | note |
| --- | --- | --- |
| easy- | [PR#](PRlink) |  |


src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
-    fetchText.asScala.foreach(item =>
-      Using.urlInputStream(new URL(item.getUrl)).foreach(src =>
-        Files.copy(src, Paths.get(bagitDir.getAbsolutePath, item.getFilename))))
-  }
-
-  def checkBagValidity(bagitDir: File)(implicit id: String): Try[Unit] = {
+  def checkBagValidity(bagitDir: File)(implicit id: String) = {
     log.debug(s"[$id] Verifying bag validity")
+
+    val fetchTexts = getFetchTxt(bagitDir).map(fetchText => fetchText.asScala).getOrElse(List()).toList
```
rvanheest: You could put the `toList` behind `asScala`: you don't need to do that on something that is already a list.

You could also leave out the `toList` and make it such that everything operates on `Seq`, the 'interface parent' of `List`. Not sure how much effort that would take, though. Maybe worth trying...

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
-    else Failure(new InvalidDepositException(id, validationResult.messagesToString))
+
+    if (fetchTextsBagStore.isEmpty) {
+      if (validationResult.isSuccess) Success(Unit)
+      else Failure(InvalidDepositException(id, validationResult.messagesToString))
+    }
+    else {
+      if (validationResult.isSuccess) log.warn(s"[$id] There is a fetch.txt file, while all the files are present in the bag.")
+
+      val otherThanMissingPayloadFilesMessages = validationResult.getSimpleMessages.listIterator().asScala.filterNot(msg =>
```
rvanheest: why you need a `listIterator` here and below?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
+        Using.urlInputStream(new URL(item.getUrl)).foreach(src =>
+          Files.copy(src, Paths.get(bagitDir.getAbsolutePath, item.getFilename))))
+    }
+  }
+
+  private def validateChecksumsFetchtextFiles(bag: Bag, fetchTexts: List[FetchTxt.FilenameSizeUrl]) (implicit id: String) = {
+    log.debug(s"[$id] Validating the checksums of those files in fetch.txt, that refer to the bag store.")
+
+    val fetchTextFiles = fetchTexts.map(item => item.getFilename)
+    val checksums = bag.getPayloadManifests.asScala.toList.flatMap(manifest => manifest.asScala).filter(fileChecksumTuple => fetchTextFiles.contains(fileChecksumTuple._1))
```
rvanheest: could you please do a pattern match on `fileChecksumTuple`?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
+    val urls = fetchTexts.map(file => file.getFilename -> file.getUrl)
```
rvanheest: if you make this into a `Map`, the `checksumMapping` gets easier to read!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
+        urls.find { case (file_b, url) => file_a == file_b }
+          .map { case (_, url) => (file_a, checksum, url) }
+            .getOrElse(throw InvalidDepositException(id, s"Checksum validation failed: missing Payload Manifest file $file_a not found in the fetch.txt."))
+        }
+
+    validateChecksums(checksumMapping)
+  }
+
+  private def validateChecksums(checksumMapping: List[(String, String, String)]) (implicit id: String) = {
+    checksumMapping.map(tuple => compareChecksumAgainstReferredBag(tuple._1,tuple._2,tuple._3)).collectResults
```
rvanheest: pattern match please!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
+
+    val fetchTextFiles = fetchTexts.map(item => item.getFilename)
+    val checksums = bag.getPayloadManifests.asScala.toList.flatMap(manifest => manifest.asScala).filter(fileChecksumTuple => fetchTextFiles.contains(fileChecksumTuple._1))
+    val urls = fetchTexts.map(file => file.getFilename -> file.getUrl)
+
+    val checksumMapping: List[(String, String, String)] =
+      checksums.map { case (file_a, checksum) =>
+        urls.find { case (file_b, url) => file_a == file_b }
+          .map { case (_, url) => (file_a, checksum, url) }
+            .getOrElse(throw InvalidDepositException(id, s"Checksum validation failed: missing Payload Manifest file $file_a not found in the fetch.txt."))
```
rvanheest: why do you want to throw the exception? why not using a `Try`/`Success`/`Failure`? Hint: if you try this, you can call `collectResults` here, followed by a `flatMap` on `validateChecksums`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
+    validateChecksums(checksumMapping)
+  }
+
+  private def validateChecksums(checksumMapping: List[(String, String, String)]) (implicit id: String) = {
+    checksumMapping.map(tuple => compareChecksumAgainstReferredBag(tuple._1,tuple._2,tuple._3)).collectResults
+  }
+
+  private def compareChecksumAgainstReferredBag(file: String, checksum: String, url: String) (implicit id: String): Try[Unit] = {
+    val referredFile = getReferredFile(url)
+    val referredBagChecksums = getReferredBagChecksums(url)
```
rvanheest, line 307: Remember that a `List[(String, String)]` can also be seen as a `Map[String, String]` if the `tuple._1` is unique! This would probably make it more efficient for you in this method!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
+        Failure(InvalidDepositException(id, s"While validating checksums, the file $referredFile not found in the referred bag."))
+    }
+  }
+
+  private def getReferredFile(url: String) = {
+    val s = url.stripPrefix(baseUrl)
+    s.substring(s.indexOf("/data/") + 1)
+  }
+
+  private def getBagUrlB(dir: File)(implicit id: String): Bag = {
```
rvanheest: what does this B stand for? typo?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -277,4 +345,19 @@ object DepositHandler {
     dr.setVerboseDescription("received successfully: " + deposit.getFilename + "; MD5: " + deposit.getMd5)
     dr
   }
+
+
+  // TO DO: RETRIEVE VIA AN INTERFACE
+  private def getReferredBagChecksums(url: String) (implicit id: String) = {
+    getBagFromDir(getReferredBagDir(url)).getPayloadManifests.asScala.toList.flatMap(manifest => manifest.asScala)
```
rvanheest: If you keep out the `toList`, you have that `Map`!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -277,4 +345,19 @@ object DepositHandler {
...
+  }
+
+  private def getReferredBagDir(url: String) (implicit id: String) = {
```
rvanheest: After playing compiler in my head, combined with the example, I can figure out what this does and what everything represents, but it would be much easier for the reader if you give better names here for `s` and `dir_X`.

src/main/scala/nl/knaw/dans/api/sword2/package.scala
```diff
@@ -30,6 +30,8 @@ package object sword2 {
    * a Servlet Init Parameter into the global scope. If there is a better way, please let me know (JvM).
    */
   var homeDir: File = null
+  var baseDir: File = null
```
rvanheest: Given the comment from @janvanmansum in the code above, wouldn't it be better to put these values somewhere else. Not sure if that's possible, but maybe give it a try? Especially since these are only used in `DepositHandler` as far as I can see...

src/test/resources/simple/not-allowed-url-bag/fetch.txt
```diff
@@ -0,0 +1 @@
+ftp://dans.knaw.nl/nl/over/organisatie-beleid/informatiemateriaal/Software_Sustainability_DANS_NLeSC_2016.pdf  555292  data/brochure/Software_Sustainability_DANS_NLeSC_2016.pdf
```
rvanheest, line 1: Can we use another file here that we ourselves have more control over? Our tests will unexpectedly fail as soon as this file gets removed or replaced or modified!

@janvanmansum maybe put this file on teasy/aeasy?

@vesaakerman have you tested this with no internet? Maybe ask @jo-pol how to do this.

rvanheest, line 1: **General remark about the test resources:** given the large number of examples/bags/revisions/etc. would it be smart to make some kind of `README.md` in `src/test/resources/` with a small description of each example/bag/revision, what it is used for and which tests make use of each instance?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -62,11 +62,13 @@ object DepositHandler {
 
   def finalizeDeposit(mimeType: String)(implicit id: String): Try[Unit] = {
```
rvanheest, line 64: **general remark on the `implicit id: String` parameters in many of these methods:** are you sure this `String` is required everywhere you put it? Can't you remove it in some instances? Not sure, but it seems a bit much...

rvanheest (changes requested):
Great job in general. See the comments in the code. Another general comment: if you chain operators in for example collection manipulation, it would be more readable if you put them on separate lines.

src/main/assembly/dist/cfg/application.properties
```diff
@@ -10,4 +10,6 @@ auth.ldap.sword-enabled-attribute-name={{ easy_sword2_auth_ldap_sword_enabled_at
 auth.ldap.sword-enabled-attribute-value={{ easy_sword2_auth_ldap_sword_enabled_attribute_value }}
 auth.single.user={{ easy_sword2_auth_single_user }}
 auth.single.password={{ easy_sword2_auth_single_password }}
-url-pattern=^https?://dans.knaw.nl/.*
+url-pattern=^https?://.*
```
janvanmansum, line 13: Make the name of the config params a bit more explicit: `bag-store.base-dir`, etc.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -62,11 +62,13 @@ object DepositHandler {
...
     log.info(s"[$id] Finalizing deposit")
+    baseDir = new File(SwordProps("basedir"))
```
janvanmansum: We need some check that this directory actually exists and is read-accessible, so we fail fast if it isn't.

src/main/scala/nl/knaw/dans/api/sword2/package.scala
janvanmansum: Second that 👍

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
janvanmansum: `fetchTexts` is a misnomer. There is only one "fetch.txt" per bag. The list is of **fetch items**.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
-
-  def checkBagValidity(bagitDir: File)(implicit id: String): Try[Unit] = {
+  def checkBagValidity(bagitDir: File)(implicit id: String) = {
     log.debug(s"[$id] Verifying bag validity")
+
+    val fetchTexts = getFetchTxt(bagitDir).map(fetchText => fetchText.asScala).getOrElse(List()).toList
+    val fetchTextsBagStore = fetchTexts.filter(_.getUrl.startsWith(baseUrl))
+    val fetchTextsOther = fetchTexts diff fetchTextsBagStore
+
+    // resolve those FetchTxt files that do not refer to the bag store
```
janvanmansum: Although I have a lot to "complain" about your code, here, I think you underestimate its clarity ;)  This comment is really not needed.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
-        Files.copy(src, Paths.get(bagitDir.getAbsolutePath, item.getFilename))))
-  }
-
-  def checkBagValidity(bagitDir: File)(implicit id: String): Try[Unit] = {
+  def checkBagValidity(bagitDir: File)(implicit id: String) = {
     log.debug(s"[$id] Verifying bag validity")
+
+    val fetchTexts = getFetchTxt(bagitDir).map(fetchText => fetchText.asScala).getOrElse(List()).toList
+    val fetchTextsBagStore = fetchTexts.filter(_.getUrl.startsWith(baseUrl))
+    val fetchTextsOther = fetchTexts diff fetchTextsBagStore
```
janvanmansum: Consider removing this temp. It is not really bad, but we use it only once and the `diff` expression really says it all.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
     val validationResult: SimpleResult = bag.verifyValid
-    if (validationResult.isSuccess) Success(Unit)
-    else Failure(new InvalidDepositException(id, validationResult.messagesToString))
+
+    if (fetchTextsBagStore.isEmpty) {
+      if (validationResult.isSuccess) Success(Unit)
+      else Failure(InvalidDepositException(id, validationResult.messagesToString))
+    }
+    else {
+      if (validationResult.isSuccess) log.warn(s"[$id] There is a fetch.txt file, while all the files are present in the bag.")
```
janvanmansum: Good point to note. Actually, isn't this an error? Is this something that could feasibly happen? If not, then it is _really_ a serious error if it does, because it means the logic of our code turns out to be unsound.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,51 +154,117 @@ object DepositHandler {
...
 
-  private def resolveFetchTxtFiles(bagitDir: File, fetchText: FetchTxt)(implicit id: String): Unit = {
-    log.debug(s"[$id] Resolving files in fetch.txt")
-    fetchText.asScala.foreach(item =>
-      Using.urlInputStream(new URL(item.getUrl)).foreach(src =>
-        Files.copy(src, Paths.get(bagitDir.getAbsolutePath, item.getFilename))))
-  }
-
-  def checkBagValidity(bagitDir: File)(implicit id: String): Try[Unit] = {
+  def checkBagValidity(bagitDir: File)(implicit id: String) = {
```
janvanmansum: Make your return types explicit. (See also my general comments.)

janvanmansum (changes requested):
Reviewed part of the code. These are my findings so far. Don't get discouraged by the number of small comments. It's all in the spirit of positive feedback. If you disagree or need clarification about some comments, don't hesitate to respond and together
we'll figure out the most sensible solution.

Here are some more general point of advice:
- Naming is incredibly important. Try to stay close to the terminology used in the BagIt specs
  and in the easy-bag-store README. If there is no established term of course we'll need to invent
  one. That is why I came up with **virtually-valid** for example.
- Provide explicit return types for class and object level methods. (For function local or anonymous 
  methods it may be overkill.) Although the compiler calculates the return types for you it is not
  a good idea to rely on that. The compiler can only check that the types are consistent, not that
  they actually make sense. For example, that is why you get functions that return `Try[Object]`; the 
  compiler sees that some function exit points return `Try[Unit]` and other `Try[List[Unit]]` and 
  generalizes that accordingly.
- Use the `_` more often in anonymous functions; reading things like xs.filter(f => f.field == 'y')
  gets tedious after a while.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: In `getBagFromDir` the implicit parameter was there for nothing.  After I removed it, I could also remove it from three other methods.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: I changed over to `Seq`, it was not difficult.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Otherwise I can't apply `filter`

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Much better indeed...

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: That is a good idea, gets indeed much easier to read!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Why not? The exception will be caught in `finalizeDeposit` method, where all the error messages are logged.  If I would make it `Try/Success/Failure`, then I would need to write more code to dig up the original error message.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Yes, much better....

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: In principle it is possible that tuple._1 is not unique (when there are more than one `manifest` file), so here we can't use `Map`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Yes, a typo.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Map not possible, because the first element not necessarily unique.

src/main/scala/nl/knaw/dans/api/sword2/package.scala
vesaakerman: I was expecting this comment:) True. I added now two implicit variables.

src/test/resources/simple/not-allowed-url-bag/fetch.txt
vesaakerman: Here we actually expect that this url will not be accepted.
But I use`https://dans.knaw.nl/nl/over/organisatie-beleid/informatiemateriaal/Software_Sustainability_DANS_NLeSC_2016.pdf` in valid bags. I am aware that this url might not remain forever, but I expect that it will not disappear in the near future.

vesaakerman: I see your point, but on the other hand I am bit hesitant to do that, knowing that this sort of documentation so easily starts to lag behind.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: I added an method to check these things.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: I think that it is a good idea. I added those return types.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Very true.  I changed now everywhere `fetchText` with `fetchItem` .

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: done

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: So as agreed on Thursday, I give now an error message.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,54 +166,157 @@ object DepositHandler {
...
-    fetchText.asScala.foreach(item =>
-      Using.urlInputStream(new URL(item.getUrl)).foreach(src =>
-        Files.copy(src, Paths.get(bagitDir.getAbsolutePath, item.getFilename))))
-  }
-
-  def checkBagValidity(bagitDir: File)(implicit id: String): Try[Unit] = {
+  def checkBagValidity(bagitDir: File)(implicit id: String, baseDir: File, baseUrl: URI) = {
     log.debug(s"[$id] Verifying bag validity")
+
+    val fetchItems = getFetchTxt(bagitDir).
```
rvanheest: I noticed a slight confusion with our code conventions here. If you break up a sequence of operators into multiple lines, you put the `.` as the first character on the new line.

``` scala
val fetchItems = getFetchTxt(bagitDir)
    .map(_.asScala)
    .getOrElse(Seq())
```

Also note the indent of the second and third line! In some cases you did not have that either.

Would you mind going through the code with this in mind?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: Nope, you don't! Try it without and you'll see that `asScala` will automatically turn it into an `Iterable`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,54 +166,157 @@ object DepositHandler {
...
+      Success(Unit)
+  }
+
+  private def validateChecksumsFetchItems(bag: Bag, fetchItems: Seq[FetchTxt.FilenameSizeUrl]) (implicit id: String, baseDir: File, baseUrl: URI): Try[Unit] = {
+    log.debug(s"[$id] Validating checksums of those files in fetch.txt, that refer to the bag store.")
+
+    val fetchItemFiles = fetchItems.map(_.getFilename)
+    val checksums = bag.getPayloadManifests.asScala.toList.
+                      flatMap(_.asScala).
+                      filter({case (file, checksum) =>
```
rvanheest: As you said, this already looks so much better indeed. Two pieces of follow-up:
1. You don't need the `()` here, now that you use `{ case ... }` (saw a similar thing in other places, btw)
2. You don't use `checksum`, so you can write the pattern match as `case (file, _)`. That way it is clear to everybody that you only use the first part of the tuple and are not interested in the second part for now...

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,54 +166,157 @@ object DepositHandler {
...
+      Success(Unit)
+  }
+
+  private def compareChecksumAgainstReferredBag(file: String, checksum: String, url: String) (implicit id: String, baseDir: File, baseUrl: URI): Try[Exception] = Try {
+    val referredFile = getReferredFile(url, baseUrl)
+    val referredBagChecksums = getReferredBagChecksums(url)
+    if (referredBagChecksums.contains(referredFile -> checksum))
+      null
+    else {
+      if (referredBagChecksums.map({case (file, _) => file}).contains(referredFile))
```
rvanheest: According to my IDEA there is an issue with variable shadowing. You have a `file` as parameter of `compareChecksumAgainstReferredBag` and also have a `file` in this lambda.

src/test/resources/simple/not-allowed-url-bag/fetch.txt
rvanheest: For a similar case in the sword2-examples project we have uploaded a couple of files to GitHub. Maybe you could do the same for this project.

rvanheest: That's what is called discipline! We ought to keep each other to the highest standards, especially regarding this kind of things. But to be honest with you, I have not the slightest clue what differences there are between the various bags in the various folders/sub-folders. Folder-names do not solve this issue completely. With some prior knowledge I understand that `invalid-url-bag` has something to do with `fetch.txt` (or maybe not?), but where can I find the difference between `example-bag` and `simple-bag` unless starting a new carrier as a detective? That's why we need this documentation on root level!

rvanheest (commented):
Not yet done with the whole review, but here are some comments in the code already.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: You didn't for `checkBagValidity`. No return type. However, if you generate it automagically, it will give you `Try[Any]`. That's because you have a `Success(Unit)` in your code. This is incorrect. Let me explain this:

`Unit` is a **type**, just like `Boolean`, `Int`, `String`, etc. `Unit` has only one value, namely `()`. The confusing bit is that you pronounce this as _unit_ as well. So, you have to write `()` rather than `Unit`.

If you change this, you will get the correct return type, which is `Try[Unit]`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: not sure what `val tempDir = new File(SwordProps("tempdir"), id)` represents, but shouldn't you do a similar check for this one?

rvanheest: Oh, wait, you only use it in `getBagDir(tempDir)` and there it turns out to be the `depositDir`. Is this a naming issue or what? `tempDir` does not really sound like the directory where the actual deposit is located...

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -78,8 +81,8 @@ object DepositHandler {
       case FailedDepositException(_, msg, cause) =>
         log.error(s"[$id] Failed deposit", cause)
         DepositProperties.set(id, "FAILED", msg, lookInTempFirst = true)
-      case cause: Throwable =>
-        log.error(s"[$id] Unexpected failure in deposit", cause)
+      case _: Throwable =>
```
rvanheest: Why did you remove the `cause`. I'd love to have this one during a debugging session based on production logs!

rvanheest: Actually, now that I think about this some more, we need to take a customer in mind here, don't we? After all, the `DepositProperties.set` texts are being seen by the depositor that submits its data via Sword. This means that we really need to look carefully at every error message that can possible occur in the `DepositProperties`! An error message like `Unexpected failure in deposit` is no good, if you ask me. What is the customer supposed to do with this? Try again, only to find a similar kind of error message? Is this because of him or the dataset he tries to deposit, or is this an error on our side? I see why we need a default case with a pattern match on `Throwable`, but maybe we can be somewhat more specific?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -121,6 +124,17 @@ object DepositHandler {
     }
   }
 
+  def checkBaseDir()(implicit id: String, baseDir: File): Try[Unit] = Try {
```
rvanheest: First of all, why would you wrap this method in a `Try` and then **throw** exceptions? Why not just use the proper way of `Success` and `Failure`?

Second, with again the customer in mind, are these error messages suitable for the customer to be read. Do they need to know that we configured our Bag Store incorrectly? I guess that's not really what we want, right? However, I don't know a proper alternative yet. Maybe something like '_contact customer support_' or something? Suggestions needed here!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,54 +166,157 @@ object DepositHandler {
...
-  def checkBagValidity(bagitDir: File)(implicit id: String): Try[Unit] = {
+  def checkBagValidity(bagitDir: File)(implicit id: String, baseDir: File, baseUrl: URI) = {
     log.debug(s"[$id] Verifying bag validity")
+
+    val fetchItems = getFetchTxt(bagitDir).
+                      map(_.asScala).
+                      getOrElse(Seq())
+    val fetchItemsBagStore = fetchItems.filter(_.getUrl.startsWith(baseUrl.toString))
+
+    resolveFetchItems(bagitDir, fetchItems diff fetchItemsBagStore)
```
rvanheest: It seems that you do a check on the existence of the files in `fetch.txt` twice: once in `resolveFetchItems` and once in `noFetchItemsAlreadyInBag`. Is that correct? And is this necessary?

rvanheest (commented):
Here are some more comments.
Some of them are taken care of in https://github.com/vesaakerman/easy-sword2/pull/1

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: there is so much yet to learn in Scala...:)

src/test/resources/simple/not-allowed-url-bag/fetch.txt
vesaakerman: That is better indeed. Then you know for sure that the file doesn't suddenly disapperar.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: The bag will be moved from the tempDir to the proper deposit directory.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Thanks for the explanation.  I had copied `Success(Unit)` from other code.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Yes, I see it now, too.  I'll rename it.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Removing the `cause` was an accident.  Now I remember that I wanted to test something in the behavior of showing the error message and had then forgotten to cancel the change.

src/main/assembly/dist/cfg/application.properties
```diff
@@ -12,4 +12,4 @@ auth.single.user={{ easy_sword2_auth_single_user }}
 auth.single.password={{ easy_sword2_auth_single_password }}
 url-pattern=^https?://.*
 bag-store.base-url={{ easy_bag_store_baseuri }}
-bag-store.base-dir={{ easy_bag_store_default_basedir }}
+bag-store.base-dir={{ easy_bag_store_basedir }}
```
rvanheest, line 15: note that with this change you also have to change dtap!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Well, I thought that it is also proper Scala to wrap a method in a `Try`.
As you have seen, I use in more places in this code exceptions with the idea that the program goes directly to the main method where the exception will be caught.  I have noticed that also in other places you prefer to have Failure rather than throw exception. I am wondering, is that a matter of taste, or do you find that throwing an exception is really bad?

vesaakerman: One idea might be to give `FailedDepositException` only when there is something wrong on our side.  And when the cause is in the deposited bag, then always give `InvalidDepositException`.  And when `FailedDepositException`, then not to send the cause to the customer, but some other kind of explanation.

rvanheest: We discussed the option to use Scala's `NonFatal(e)` error catching.

PR #66, opened 2016-10-20
Improve client's error messages

rvanheest:
fixes EASY-1135
#### When applied it will
- [x] give clear error messages to the client when its input is invalid
- [x] give a generic error message when an internal error occurs (server-side)
- [x] take proper care of `CompositeException`s
- [x] remove the `FailedDepositException`

`DepositHandler.scala`
- [x] test the whole stuff

Give as many invalid inputs as you can possible think of and see what the application gives you back! Think 'invalid URLs', 'not allowed URLs', 'invalid checksums', 'fetch items refer to paths that already exist (same file)', 'fetch item overrides file that does already exist (other file)'.

You can use [easy-sword2-dans-examples](https://github.com/DANS-KNAW/easy-sword2-dans-examples) to submit your test bags!

| repo | PR |
| --- | --- |
| easy-dtap | [PR#46](https://github.com/DANS-KNAW/easy-dtap/pull/46) |

### DIFF
--- a/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
+++ b/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
@@ -94,11 +94,8 @@ object DepositHandler {
       case InvalidDepositException(_, msg, cause) =>
         log.error(s"[$id] Invalid deposit", cause)
         DepositProperties.set(id, "INVALID", msg, lookInTempFirst = true)
-      case FailedDepositException(_, msg, cause) =>
-        log.error(s"[$id] Failed deposit", cause)
-        DepositProperties.set(id, "FAILED", genericErrorMessage, lookInTempFirst = true)
-      case NonFatal(e)  =>
-        log.error(s"[$id] Unexpected failure in deposit", e)
+      case NonFatal(e) =>
+        log.error(s"[$id] Internal failure in deposit", e)
         DepositProperties.set(id, "FAILED", genericErrorMessage, lookInTempFirst = true)
     }
   }
@@ -142,8 +139,8 @@ object DepositHandler {
   }
 
   def checkBagStoreBaseDir()(implicit id: String, baseDir: File): Try[Unit] = {
-    if (!baseDir.exists) Failure(FailedDepositException(id, s"Bag store base directory ${baseDir.getAbsolutePath} doesn't exist"))
-    else if (!baseDir.canRead) Failure(FailedDepositException(id, s"Bag store base directory ${baseDir.getAbsolutePath} is not readable"))
+    if (!baseDir.exists) Failure(new IOException(s"Bag store base directory ${baseDir.getAbsolutePath} doesn't exist"))
+    else if (!baseDir.canRead) Failure(new IOException(s"Bag store base directory ${baseDir.getAbsolutePath} is not readable"))
     else Success(())
   }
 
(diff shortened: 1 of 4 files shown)


rvanheest:
I'm thinking of removing the `FailedDepositException` completely, as it is only used in `checkBagStoreBaseDir` to check whether the bag-store exists and is readable. To me these are more of an `IOException`. Besides that, we are planning to make the bag-store stand-alone (microservice), so then this will be removed anyway...

LMKWYT!

rvanheest:
Finished. Please review!

rvanheest:
**A point for discussion:**

In the case an `InvalidDepositException` exception occurs (e.g. when the user has submitted an invalid deposit), do we really need to log the whole stack trace as an error? Surely, it is useful for the user to see what's going wrong, such that he/she can fix it and try again (**note:** he/she doesn't get to see any stacktrace!), but do _we_ need to see both the error message and its full stacktrace? If you'd ask me, it would be enough to log`log.error(s"[$id] Invalid deposit: $msg")` rather than the current `log.error(s"[$id] Invalid deposit", cause)`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -82,12 +94,9 @@ object DepositHandler {
...
         log.error(s"[$id] Invalid deposit", cause)
         DepositProperties.set(id, "INVALID", msg, lookInTempFirst = true)
-      case FailedDepositException(_, msg, cause) =>
-        log.error(s"[$id] Failed deposit", cause)
-        DepositProperties.set(id, "FAILED", msg, lookInTempFirst = true)
-      case NonFatal(e)  =>
-        log.error(s"[$id] Unexpected failure in deposit", e)
-        DepositProperties.set(id, "FAILED", "Unexpected failure in deposit", lookInTempFirst = true)
+      case NonFatal(e) =>
+        log.error(s"[$id] Internal failure in deposit", e)
```
janvanmansum, line 98: "... in deposit _service_" to make it more explicit, otherwise it half sounds like we are still blaming the client.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -97,40 +106,41 @@ object DepositHandler {
     def getSequenceNumber(f: File): Int =
       try {
         f.getName.split('.').last.toInt
-      } catch {
+      }
```
janvanmansum, line 109: Not part of this change, but actually I think catching all `Throwable`s below is also a bug. We should catch the specific error that results from having something else than an integer. (By the way we do not even check that it is positive!) Other exceptions should lead to an "internal error".

janvanmansum (approved):
Looks quite good, nice consistent style!

During testing I found I hard to create an "internal error". Even when messing up the `deposit.properties` the service just leaves out the properties it cannot find. I think it would be better to have those conditions lead to a FailedDeposit, as the client code probably depends on being able to retrieve those items. 

This was probably not introduced by this changed, so I'll merge it and create a new issue for "failing fast".

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: It doesn't get prettier, but here's a shot at making it better (see #67)

PR #72, opened 2016-11-14
EASY-1148, delete fetched items from fetch.txt ->

janvanmansum:
fixes EASY-1148

#### When applied it will
* Now deleting all items that were fetched to the bag from fetch.txt. If fetch.txt is empty
after that, it is deleted.

At [the new function](https://github.com/DANS-KNAW/easy-sword2/compare/master...janvanmansum:EASY-1148?expand=1#diff-ad17b49d50f1b93808fb326e8f79d0d8R277)

Use the fetch-example from `easy-sword2-dans-examples`, deposit is and verify that the fetch.txt is deleted, or run it with the examples themselves and user mendeltest and see that archiving to archival storage does not fail any longer.

- [x] Finish refactoring getBagFromDir so that the new version, returning a Try is used everywhere
- [x] Test on `deasy`


janvanmansum:
@DANS-KNAW/easy for review.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -263,14 +264,24 @@ object DepositHandler {
...
+      _ <- resolveFetchItems(bagitDir, itemsToResolve)
+      _ <- removeFromFetchTxt(bagitDir, itemsToResolve)
       bag = getBagFromDir(bagitDir)
       validationResult = bag.verifyValid
       _ <- handleValidationResult(bag, validationResult, fetchItemsInBagStore)
     } yield ()
   }
 
+  private def removeFromFetchTxt(bagDir: File, items: Seq[FetchTxt.FilenameSizeUrl]): Try[Unit] = Try {
+    getFetchTxt(bagDir).foreach { case fetchTxt =>
```
rvanheest: I think it is better to use a `map` here:

``` scala
getFetchTxt(bagDir)
    .map(fetchTxt => Try {
        items.foreach(fetchTxt.remove)
        if (fetchTxt.isEmpty) new File(bagDir, "fetch.txt").delete()
        getBagFromDir(bagDir).write(new FileSystemWriter(bagFactory), bagDir)
    })
    .map(_ => ())
    .getOrElse(Success(()))
```

janvanmansum: Why? It looks more complicated. What is the advantage?

rvanheest: Seems more obvious to me. In general I'm always careful with the use of `foreach`, as it does not return anything. In the `items.foreach(fetchTxt.remove)` it is a logical choice to do so, but to me it is not at the `getFetchTxt.foreach`. I'd say you want a `Try` back, so you use `map`.

Also, the `getOrElse` makes it more explicitly clear that it is still OK if the `getFetchTxt` returns a `None`. For that to work you also require a `map`.

Sidenote: you could also merge the two `map`s and just return a `()` at the end of the first `map`. The reason I choose for the approach with the second `map` is purely esthetic.

Since each of the three parts of the first `map` are functions that might throw an exception (welcome to the stupidly impure world of Java!), we require this `Try` around it to catch all exceptions and return in `Failure` rather than a `Success` if an exception is thrown.

jo-pol: What if the delete fails: Isn't that notified with a boolean result rather than an exception?

rvanheest: it's indicated with both: https://docs.oracle.com/javase/8/docs/api/java/io/File.html#delete--

janvanmansum: The API is a bit vague I think. Surely you cannot do both at the same time. A few tests seem to indicate that deleting file that doesn't exist of for which you have no permissions to delete it are handled by returning `false` rather than throwing an exception, so @jo-pol 's point is valid.

rvanheest (approved):
Looks good to me!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -267,21 +269,31 @@ object DepositHandler {
...
+          val completer = new TagManifestCompleter(bagFactory)
+          completer.setTagManifestAlgorithm(a)
+          completer complete bag
+        })
+        val writer = new FileSystemWriter(bagFactory)
+        writer.setTagFilesOnly(true)
+        bag.write(writer, bagDir)
+    }
 
+  private def getBagFromDir2(bagDir: File): Try[Bag] =  Try {
```
rvanheest: Ehm,... numbers in function names? Did you forget to refactor something?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -267,21 +269,31 @@ object DepositHandler {
...
 
-  private def removeFromFetchTxt(bagDir: File, items: Seq[FetchTxt.FilenameSizeUrl]): Try[Unit] =
-    getFetchTxt(bagDir).map(fetchTxt => Try {
-      items.foreach(fetchTxt.remove)
-      if (fetchTxt.isEmpty && !new File(bagDir, "fetch.txt").delete()) throw new IllegalStateException("Unable to remove empty fetch.txt")
-      getBagFromDir(bagDir).write(new FileSystemWriter(bagFactory), bagDir)
-    }.map(_ => ()))
-    .getOrElse(Success(()))
+  def pruneFetchTxt(bagDir: File, items: Seq[FetchTxt.FilenameSizeUrl]): Try[Unit] =
+    getBagFromDir2(bagDir).map {
```
rvanheest: What I see you do often is these pattern matches. Can you explain why you need these. Is there to you a difference between `.map { case bag => ... }` and `.map(bag => ...)`? Because you use the former here and the latter at line 284.

I'd say, only use pattern matching when needed (a.k.a. you distinguish between some inputs or write a partial function)!

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: Other question: is it necessary to wrap `getBagFromDir` in a `Try`? If so, why haven't we done that in all other uses of `getBagFromDir`?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -267,21 +269,31 @@ object DepositHandler {
...
+      case bag =>
+        bag.getFetchTxt.removeAll(items.asJava)
```
rvanheest: You did not incorporate the possibility that `bag.getFetchTxt` is `null` (which it will, if you look into the code!)

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -267,21 +269,31 @@ object DepositHandler {
...
+          completer complete bag
+        })
+        val writer = new FileSystemWriter(bagFactory)
+        writer.setTagFilesOnly(true)
+        bag.write(writer, bagDir)
+    }
 
+  private def getBagFromDir2(bagDir: File): Try[Bag] =  Try {
+    getBagFromDir(bagDir)
+  }
```
rvanheest, line 296: Here is an alternative implementation (tested in your branch, passes all tests) that takes into account the fact that `getFetchTxt` can be null and does not require a `getBagFromDir2`

``` scala
def pruneFetchTxt(bagDir: File, items: Seq[FetchTxt.FilenameSizeUrl]): Try[Unit] = {
  val bag = getBagFromDir(bagDir)

  def cleanFetchTxt(fetchTxt: FetchTxt): Try[Unit] = Try {
    fetchTxt.removeAll(items.asJava)
    if (fetchTxt.isEmpty)
      bag.removeBagFile(bag.getBagConstants.getFetchTxt)

    bag.getTagManifests
      .asScala
      .map(_.getAlgorithm)
      .foreach(algorithm => {
        val completer = new TagManifestCompleter(bagFactory)
        completer.setTagManifestAlgorithm(algorithm)
        completer.complete(bag)
      })
    val writer = new FileSystemWriter(bagFactory)
    writer.setTagFilesOnly(true)
    writer.write(bag, bagDir)
  }

  Option(bag.getFetchTxt)
    .map(cleanFetchTxt)
    .getOrElse(Success(()))
}
```

src/test/scala/nl/knaw/dans/api/sword2/PruneFetchTxtSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+package nl.knaw.dans.api.sword2
+
+import java.io.File
+
+import gov.loc.repository.bagit.FetchTxt.FilenameSizeUrl
+
+import scala.io.{Codec, Source}
+
+class PruneFetchTxtSpec extends Sword2Fixture {
+  implicit val codec =Codec.UTF8
```
rvanheest, line 26: hit the spacebar one more time!

src/test/scala/nl/knaw/dans/api/sword2/PruneFetchTxtSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  it should "remove fetch.txt if it is left empty after pruning" in {
+    copyToTargetBagDir(TEST_BAG)
+    DepositHandler.pruneFetchTxt(targetBagDir, Seq(FETCH_ITEM_1, FETCH_ITEM_2))
+    FETCH_TXT shouldNot exist
+  }
+
+  it should "update checksum for fetch.txt in tagmanifest if fetch.txt is changed" in {
+    copyToTargetBagDir(TEST_BAG)
+    getTagManifestSrc should include regex(s"""${CHECKSUM_FETCH_TXT_WITH_BOTH_FETCH_ITEMS}\\s+fetch.txt""") // Double-check preconditions
```
rvanheest: You don't need the `(` and `)` here and below, according to IntelliJ

rvanheest: neither do you need the `{` and `}` here and below

src/test/scala/nl/knaw/dans/api/sword2/PruneFetchTxtSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+    DepositHandler.pruneFetchTxt(targetBagDir, Seq(new FilenameSizeUrl("data/file/in/bag", 0L, "http://some/url")))
+    getTagManifestSrc should include regex(s"""${CHECKSUM_FETCH_TXT_WITH_ONLY_FETCH_ITEM_2}\\s+fetch.txt""") // Checksum of fetch.txt has changed because one item was deleted
+  }
+
+  it should "remove fetch.txt from tagmanifest if fetch.txt is removed" in {
+    copyToTargetBagDir(TEST_BAG)
+    DepositHandler.pruneFetchTxt(targetBagDir,
+      Seq(new FilenameSizeUrl("data/file/in/bag", 0L, "http://some/url"),
+        new FilenameSizeUrl("data/other/file/in/bag", 0L, "http://some/other/url")))
+    getTagManifestSrc shouldNot include("fetch.txt")
```
rvanheest: what about `getTagManifestSrc should not include "fetch.txt"`?

rvanheest (changes requested):
Your tests indeed reveal some bugs in the earlier version! But here is another test: "_your bag does not have a `fetch.txt`_". Some of the comments below go more into that case and provide an alternative implementation.

src/test/scala/nl/knaw/dans/api/sword2/PruneFetchTxtSpec.scala
janvanmansum: 👍

src/test/scala/nl/knaw/dans/api/sword2/PruneFetchTxtSpec.scala
janvanmansum: 👍

janvanmansum:
@DANS-KNAW/easy, please have another look.
- Refactored `getBagFromDir` to return a Try
- Removed `getFetchFile` (using `getBagFromDir.map(_.getFetchFile)`)

rvanheest:
1. Looks like an interesting refactoring! Improved quite a lot. Also I see now why you wrapped `getBagFromDir` in a `Try`. It's indeed better this way!
2. I noticed that `getFetchFile` not only requires the `getBagFromDir.map(_.getFetchFile)`, but also don't forget to have the `.filter(_ != null)`! Maybe that extra method is better here? It would either have signature `getFetchTxt(bagDir: File): Try[FetchTxt]` or `getFetchTxt(bagDir: File): Try[Option[FetchTxt]]`.
3. `pruneFetchTxt` still uses the getFetchFile without the null-check. This is also related to the unit test with the scenario that the bag has no `fetch.txt` file. In my opinion this should just succeed rather than fail. If there is no `fetch.txt`, then there is nothing to prune, so, fine, let's celebrate a cheap `Success`! I'd say, keep the `Failure`s for the real failing cases!
4. I notice multiple calls to `getBagFromDir`, some of which actually want the `fetch.txt`, others want something different from it. Would it be possible to only once get this `Bag` and then pass it on as a parameter. The same probably holds for `fetch.txt`...
5. We now have (and always had) `getBagDir` and `getBagFromDir`. Could we make these names less confusing by for example renaming `getBagFromDir` to `getBag`?
6. I'm not quite certain about the new additions to `getReferredBagChecksums`: if the `getBagFromDir` gives a `Failure`, do you really want to change that into a `Seq.empty`? Does not seem obvious to me, but maybe you have reasons for that?

rvanheest (approved):
Looks good to me!

PR #73, opened 2016-11-21
EASY-1145  Using bag-store in easy-sword2 configurable

vesaakerman:
fixes EASY-1145

#### When applied it will
* it will check that either both bag-store `base-dir` and `base-url` are given, or that neither of them is given
* it will skip checking existence of the `bag-store ` `base-dir` if `base-dir` and `base-url` are absent
* if both bag-store base-dir and base-url are absent, all fetch-txt items will be fetched (also those referring to the bag-store)

* set bag-store.base-dir and bag-store.base-uri to an empty string, or just one of them to an empty string.
* send a bag to sword2 and check the status

@DANS-KNAW/easy

### DIFF
--- a/src/main/scala/nl/knaw/dans/api/sword2/CollectionDepositManagerImpl.scala
+++ b/src/main/scala/nl/knaw/dans/api/sword2/CollectionDepositManagerImpl.scala
@@ -16,6 +16,7 @@
 package nl.knaw.dans.api.sword2
 
 import nl.knaw.dans.api.sword2.DepositHandler._
+import nl.knaw.dans.api.sword2.State._
 import org.apache.commons.lang.StringUtils._
 import org.swordapp.server._
 
@@ -47,7 +48,7 @@ class CollectionDepositManagerImpl extends CollectionDepositManager {
   private def setDepositStateToDraft(id: String, userId: String)(implicit settings: Settings): Try[Unit] =
     DepositProperties.set(
       id = id,
-      stateLabel = "DRAFT",
+      stateLabel = DRAFT,
       stateDescription = "Deposit is open for additional data",
       userId = Some(userId),
       lookInTempFirst = true)
--- a/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
+++ b/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
@@ -81,11 +81,10 @@ object DepositHandler {
 
   def finalizeDeposit(mimeType: String)(implicit settings: Settings, id: String): Try[Unit] = {
     log.info(s"[$id] Finalizing deposit")
-    implicit val bagStoreBase = BagStoreBase(settings.bagStoreBaseDir, settings.bagStoreBaseUrl)
+    implicit val bagStoreSettings = settings.bagStoreSettings
     val tempDir = new File(settings.tempDir, id)
 
     val result = for {
-      _        <- setBagStoreAwareness
       _        <- checkBagStoreBaseDir
       _        <- extractBag(mimeType)
       bagDir   <- getBagDir(tempDir)
@@ -102,15 +101,6 @@ object DepositHandler {
       case NonFatal(e) =>
         log.error(s"[$id] Internal failure in deposit service", e)
         DepositProperties.set(id, FAILED, genericErrorMessage, lookInTempFirst = true)
-    }
-  }
-
-  def setBagStoreAwareness(implicit bagStoreBase: BagStoreBase): Try[Unit] = {
-    bagStoreBase match {
-      case bs if bs.baseDir.trim.isEmpty && bs.baseUrl.trim.isEmpty => Success(bagStoreBase.isBagStoreAware = false)
-      case bs if bs.baseDir.trim.isEmpty => Failure(new IllegalArgumentException("Only bag store base-url given, bag store base-directory missing"))
-      case bs if bs.baseUrl.trim.isEmpty => Failure(new IllegalArgumentException("Only bag store base-directory given, bag store base-url missing"))
-      case _ => Success(bagStoreBase.isBagStoreAware = true)
     }
   }
 
@@ -160,9 +150,9 @@ object DepositHandler {
     }
   }
 
-  def checkBagStoreBaseDir()(implicit id: String, bs: BagStoreBase): Try[Unit] = {
-    if (bs.isBagStoreAware) {
-      val baseDir = new File(bs.baseDir)
+  def checkBagStoreBaseDir()(implicit id: String, bs: Option[BagStoreSettings]): Try[Unit] = {
+    if (bs.nonEmpty) {
+      val baseDir = new File(bs.get.baseDir)
       if (!baseDir.exists) Failure(new IOException(s"Bag store base directory ${baseDir.getAbsolutePath} doesn't exist"))
       else if (!baseDir.canRead) Failure(new IOException(s"Bag store base directory ${baseDir.getAbsolutePath} is not readable"))
       else Success(())
@@ -240,7 +230,7 @@ object DepositHandler {
     } yield ()
   }
 
-  def checkBagVirtualValidity(bagDir: File)(implicit id: String, bs: BagStoreBase): Try[Unit] = {
+  def checkBagVirtualValidity(bagDir: File)(implicit id: String, bs: Option[BagStoreSettings]): Try[Unit] = {
     log.debug(s"[$id] Verifying bag validity")
 
     def handleValidationResult(bag: Bag, validationResult: SimpleResult, fetchItemsInBagStore: Seq[FilenameSizeUrl]): Try[Unit] = {
@@ -272,7 +262,7 @@ object DepositHandler {
     }
 
     val fetchItems = getFetchTxt(bagDir).map(_.asScala).getOrElse(Seq())
-    val (fetchItemsInBagStore, itemsToResolve) = fetchItems.partition(bs.isBagStoreAware && _.getUrl.startsWith(bs.baseUrl))
+    val (fetchItemsInBagStore, itemsToResolve) = fetchItems.partition(bs.nonEmpty && _.getUrl.startsWith(bs.get.baseUrl))
     for {
       _ <- resolveFetchItems(bagDir, itemsToResolve)
       _ <- if(itemsToResolve.isEmpty) Success(()) else pruneFetchTxt(bagDir, itemsToResolve)
@@ -343,7 +333,7 @@ object DepositHandler {
       Success(())
   }
 
-  private def validateChecksumsFetchItems(bag: Bag, fetchItems: Seq[FetchTxt.FilenameSizeUrl])(implicit id: String, bs: BagStoreBase): Try[Unit] = {
+  private def validateChecksumsFetchItems(bag: Bag, fetchItems: Seq[FetchTxt.FilenameSizeUrl])(implicit id: String, bs: Option[BagStoreSettings]): Try[Unit] = {
     log.debug(s"[$id] Validating checksums of those files in fetch.txt, that refer to the bag store.")
 
     val fetchItemFiles = fetchItems.map(_.getFilename)
@@ -356,7 +346,7 @@ object DepositHandler {
     validateChecksums(checksumMapping)
   }
 
-  private def validateChecksums(checksumMapping: Seq[(String, String, String)])(implicit id: String, bs: BagStoreBase): Try[Unit] = {
+  private def validateChecksums(checksumMapping: Seq[(String, String, String)])(implicit id: String, bs: Option[BagStoreSettings]): Try[Unit] = {
     checksumMapping
       .map {
         case (file, checksum, url) => compareChecksumAgainstReferredBag(file, checksum, url)
@@ -367,8 +357,8 @@ object DepositHandler {
       }
   }
 
-  private def compareChecksumAgainstReferredBag(file: String, checksum: String, url: String)(implicit id: String, bs: BagStoreBase): Try[Unit] = {
-    val referredFile = getReferredFile(url, bs.baseUrl)
+  private def compareChecksumAgainstReferredBag(file: String, checksum: String, url: String)(implicit id: String, bs: Option[BagStoreSettings]): Try[Unit] = {
+    val referredFile = getReferredFile(url, bs.get.baseUrl)
     getReferredBagChecksums(url).flatMap(seq => {
       if (seq.contains(referredFile -> checksum))
         Success(())
@@ -458,20 +448,20 @@ object DepositHandler {
 
 
   // TODO: RETRIEVE VIA AN INTERFACE
-  private def getReferredBagChecksums(url: String)(implicit bs: BagStoreBase): Try[Seq[(String, String)]] =
+  private def getReferredBagChecksums(url: String)(implicit bs: Option[BagStoreSettings]): Try[Seq[(String, String)]] =
     getBag(getReferredBagDir(url)).map(bag => {
       bag.getPayloadManifests
         .asScala
         .flatMap(_.asScala)
     })
 
-  private def getReferredBagDir(url: String)(implicit bs: BagStoreBase): File = {
+  private def getReferredBagDir(url: String)(implicit bs: Option[BagStoreSettings]): File = {
     //  http://deasy.dans.knaw.nl/aips/31aef203-55ed-4b1f-81f6-b9f67f324c87.2/data/x -> 31/aef20355ed4b1f81f6b9f67f324c87/2
-    val Array(uuid, version) = url.stripPrefix(bs.baseUrl)
+    val Array(uuid, version) = url.stripPrefix(bs.get.baseUrl)
       .split("/data").head.replaceAll("-", "")
       .split("\\.")
     val (topDir, uuidDir) = uuid.splitAt(3)
 
-    getFile(bs.baseDir, topDir, uuidDir, version)
+    getFile(bs.get.baseDir, topDir, uuidDir, version)
   }
 }
--- a/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
+++ b/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
@@ -249,7 +249,7 @@ object DepositHandler {
             if (missingFilesNotInFetchText.isEmpty)
               noFetchItemsAlreadyInBag(bagDir, itemsFromBagStore)
                 .flatMap(_ => bagStoreSettings.map(implicit bs => validateChecksumsFetchItems(bag, itemsFromBagStore))
-                  .getOrElse(Failure(new NoSuchElementException("BagStore is not configured"))))
+                  .getOrElse(Failure(new NoSuchElementException("BagStore is not configured - SOMETHING WENT WRONG, YOU SHOULD NEVER REACH THIS PART OF CODE!"))))
             else
               Failure(InvalidDepositException(id, s"Missing payload files not in the fetch.txt: ${missingFilesNotInFetchText.mkString}."))
           }
@@ -259,7 +259,7 @@ object DepositHandler {
     }
 
     val fetchItems = getFetchTxt(bagDir).map(_.asScala).getOrElse(Seq())
-    val (fetchItemsInBagStore, itemsToResolve) = bagStoreSettings.map(bs => fetchItems.partition(_.getUrl.startsWith(bs.baseUrl))).getOrElse((Seq.empty, fetchItems))
+    val (fetchItemsInBagStore, itemsToResolve) = fetchItems.partition(bagStoreSettings.nonEmpty && _.getUrl.startsWith(bagStoreSettings.get.baseUrl))
     for {
       _ <- resolveFetchItems(bagDir, itemsToResolve)
       _ <- if(itemsToResolve.isEmpty) Success(()) else pruneFetchTxt(bagDir, itemsToResolve)
--- a/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
+++ b/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
@@ -23,13 +23,13 @@ import java.util.Collections
 import java.util.regex.Pattern
 
 import gov.loc.repository.bagit.FetchTxt.FilenameSizeUrl
-import gov.loc.repository.bagit.Manifest.Algorithm
 import gov.loc.repository.bagit.transformer.impl.TagManifestCompleter
 import gov.loc.repository.bagit.utilities.SimpleResult
 import gov.loc.repository.bagit.verify.CompleteVerifier
 import gov.loc.repository.bagit.writer.impl.FileSystemWriter
 import gov.loc.repository.bagit.{Bag, BagFactory, FetchTxt}
 import net.lingala.zip4j.core.ZipFile
+import nl.knaw.dans.api.sword2.State._
 import nl.knaw.dans.lib.error.{CompositeException, TraversableTryExtensions}
 import org.apache.abdera.i18n.iri.IRI
 import org.apache.commons.codec.digest.DigestUtils
@@ -92,17 +92,17 @@ object DepositHandler {
       bagDir   <- getBagDir(tempDir)
       _        <- checkFetchItemUrls(bagDir, settings.urlPattern)
       _        <- checkBagVirtualValidity(bagDir)
-      _        <- DepositProperties.set(id, "SUBMITTED", "Deposit is valid and ready for post-submission processing", lookInTempFirst = true)
+      _        <- DepositProperties.set(id, SUBMITTED, "Deposit is valid and ready for post-submission processing", lookInTempFirst = true)
       dataDir  <- moveBagToStorage()
     } yield ()
 
     result.recover {
       case InvalidDepositException(_, msg, cause) =>
         log.error(s"[$id] Invalid deposit", cause)
-        DepositProperties.set(id, "INVALID", msg, lookInTempFirst = true)
+        DepositProperties.set(id, INVALID, msg, lookInTempFirst = true)
       case NonFatal(e) =>
         log.error(s"[$id] Internal failure in deposit service", e)
-        DepositProperties.set(id, "FAILED", genericErrorMessage, lookInTempFirst = true)
+        DepositProperties.set(id, FAILED, genericErrorMessage, lookInTempFirst = true)
     }
   }
 
@@ -181,7 +181,7 @@ object DepositHandler {
   def handleDepositAsync(deposit: Deposit)(implicit settings: Settings, id: String): Try[Unit] = Try {
     if (!deposit.isInProgress) {
       log.info(s"[$id] Scheduling deposit to be finalized")
-      DepositProperties.set(id, "FINALIZING", "Deposit is being reassembled and validated", lookInTempFirst = true)
+      DepositProperties.set(id, FINALIZING, "Deposit is being reassembled and validated", lookInTempFirst = true)
       depositProcessingStream.onNext((id, deposit))
     } else {
       log.info(s"[$id] Received continuing deposit: ${deposit.getFilename}")
--- a/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
+++ b/src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
@@ -16,7 +16,7 @@
 package nl.knaw.dans.api.sword2
 
 import java.io.{File, IOException}
-import java.net.{MalformedURLException, URI, URL, UnknownHostException}
+import java.net.{MalformedURLException, URL, UnknownHostException}
 import java.nio.file._
 import java.nio.file.attribute.{BasicFileAttributes, PosixFilePermissions}
 import java.util.Collections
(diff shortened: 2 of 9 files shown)


src/main/scala/nl/knaw/dans/api/sword2/CollectionDepositManagerImpl.scala
```diff
@@ -47,7 +48,7 @@ class CollectionDepositManagerImpl extends CollectionDepositManager {
   private def setDepositStateToDraft(id: String, userId: String)(implicit settings: Settings): Try[Unit] =
     DepositProperties.set(
       id = id,
-      stateLabel = "DRAFT",
+      stateLabel = DRAFT,
```
rvanheest, line 51: 👍  for your enums!

src/test/scala/nl/knaw/dans/api/sword2/ResolveFetchItemsSpec.scala
```diff
@@ -109,9 +109,37 @@ class ResolveFetchItemsSpec extends Sword2Fixture with BagStoreFixture {
...
+    DepositHandler.checkBagVirtualValidity(targetBagDir) shouldBe a[Success[_]]
+  }
+
+  it should "result in a Failure when both bag-store base-dir and base-uri are not given, and there are fetch.txt references to the bag-store"  in {
+    implicit val bagStoreBase = BagStoreBase("", "")
+    copyToTargetBagDir(SIMPLE_SEQUENCE_B)
+    DepositHandler.setBagStoreAwareness shouldBe a[Success[_]]
+    DepositHandler.checkBagStoreBaseDir() shouldBe a[Success[_]]
+    DepositHandler.checkFetchItemUrls(targetBagDir, urlPattern) shouldBe a[Success[_]]
+    DepositHandler.checkBagVirtualValidity(targetBagDir) shouldBe a[Failure[_]]
```
jo-pol: Please check details of the failure: type of exception class and/or some essential content of the message.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -228,11 +240,11 @@ object DepositHandler {
     } yield ()
   }
 
-  def checkBagVirtualValidity(bagDir: File)(implicit id: String, bagStoreBaseDir: File, bagStoreBaseUri: URI): Try[Unit] = {
+  def checkBagVirtualValidity(bagDir: File)(implicit id: String, bs: BagStoreBase): Try[Unit] = {
     log.debug(s"[$id] Verifying bag validity")
 
     val fetchItems = getFetchTxt(bagDir).map(_.asScala).getOrElse(Seq())
-    val fetchItemsInBagStore = fetchItems.filter(_.getUrl.startsWith(bagStoreBaseUri.toString))
+    val fetchItemsInBagStore = fetchItems.filter(bs.isBagStoreAware && _.getUrl.startsWith(bs.baseUrl))
```
rvanheest: remove these lines here and replace line 277 (`val itemsToResolve` etc.) with the following:

```scala
val fetchItems = getFetchTxt(bagDir).map(_.asScala).getOrElse(Seq())
val (fetchItemsInBagStore, itemsToResolve) = fetchItems.partition(_.getUrl.startsWith(bs.baseUrl))

if (fetchItemsInBagStore.nonEmpty && !bs.isBagStoreAware)
  Failure(new IllegalArgumentException("blabla"))
else
```

also indent the for-comprehension to be in the else-branch

src/test/scala/nl/knaw/dans/api/sword2/ResolveFetchItemsSpec.scala
rvanheest: We experimented a bit with this after we found that this takes quite a long time for the test to succeed (because of the downloads when the bag-store does not exists, see previous comment):

```scala
val validity = DepositHandler.checkBagVirtualValidity(targetBagDir)

validity shouldBe a[Failure[_]]
    
type expectedException = IllegalArgumentException

val theException = the [expectedException] thrownBy validity.get
theException.getMessage shouldBe "blabla"

a [expectedException] should be thrownBy validity.get
```

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -152,9 +160,13 @@ object DepositHandler {
     }
   }
 
-  def checkBagStoreBaseDir()(implicit id: String, baseDir: File): Try[Unit] = {
-    if (!baseDir.exists) Failure(new IOException(s"Bag store base directory ${baseDir.getAbsolutePath} doesn't exist"))
-    else if (!baseDir.canRead) Failure(new IOException(s"Bag store base directory ${baseDir.getAbsolutePath} is not readable"))
+  def checkBagStoreBaseDir()(implicit id: String, bs: BagStoreBase): Try[Unit] = {
+    if (bs.isBagStoreAware){
```
rvanheest: hit the spacebar one more time before the `{` 😉

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -23,13 +23,13 @@ import java.util.Collections
...
 
 import gov.loc.repository.bagit.FetchTxt.FilenameSizeUrl
-import gov.loc.repository.bagit.Manifest.Algorithm
 import gov.loc.repository.bagit.transformer.impl.TagManifestCompleter
 import gov.loc.repository.bagit.utilities.SimpleResult
 import gov.loc.repository.bagit.verify.CompleteVerifier
 import gov.loc.repository.bagit.writer.impl.FileSystemWriter
 import gov.loc.repository.bagit.{Bag, BagFactory, FetchTxt}
 import net.lingala.zip4j.core.ZipFile
+import nl.knaw.dans.api.sword2.State._
```
rvanheest, line 32: clean up the imports! `URI` is no longer needed.

rvanheest (commented):
Here are the point/code-samples from our earlier discussion, as well as some other comments

src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
```diff
@@ -25,9 +26,9 @@ import scala.util.Try
 
 object DepositProperties {
   val log = LoggerFactory.getLogger(getClass)
-  case class State(label: String, description: String, timeStamp: String)
+  case class Statement(label: String, description: String, timeStamp: String)
```
janvanmansum: Okay, I can see the problem with the name clash, but this isn't a proper solution either. Please, try to solve the deeper problem, not just the name clash, or everything will just get more confusing.

janvanmansum (changes requested):
* If you take a step and look at your code. Why do you do analysis of the bag store settings
  in DepositHandler? It makes much more sense to do that were it is already done for other
  cases, namely in  `ServiceInitializer.readSettings`
* Also, why not put `BagStoreBase` inside the settings as an `Option[BagStoreBase]`
  (or `BagStoreSettings` would be a better name, actually)? By making it an `Option` you
  don't need the boolean field `isBagStoreAware` anymore and can do the conditional 
  handling the Scala way. (The current way is more C-like, which is fine ... if you are
  programming in C ;)

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -160,9 +150,9 @@ object DepositHandler {
     }
   }
 
-  def checkBagStoreBaseDir()(implicit id: String, bs: BagStoreBase): Try[Unit] = {
-    if (bs.isBagStoreAware) {
-      val baseDir = new File(bs.baseDir)
+  def checkBagStoreBaseDir()(implicit id: String, bs: Option[BagStoreSettings]): Try[Unit] = {
+    if (bs.nonEmpty) {
```
rvanheest: use `map` and `getOrElse` here.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -81,11 +81,10 @@ object DepositHandler {
 
   def finalizeDeposit(mimeType: String)(implicit settings: Settings, id: String): Try[Unit] = {
     log.info(s"[$id] Finalizing deposit")
-    implicit val bagStoreBase = BagStoreBase(settings.bagStoreBaseDir, settings.bagStoreBaseUrl)
+    implicit val bagStoreSettings = settings.bagStoreSettings
```
rvanheest, line 83: Don't think you need this implicit here. Just pass on the `settings` object to the subsequent calls and get you `Option[BagStoreSettings]` there.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -272,7 +262,7 @@ object DepositHandler {
     }
 
     val fetchItems = getFetchTxt(bagDir).map(_.asScala).getOrElse(Seq())
-    val (fetchItemsInBagStore, itemsToResolve) = fetchItems.partition(bs.isBagStoreAware && _.getUrl.startsWith(bs.baseUrl))
+    val (fetchItemsInBagStore, itemsToResolve) = fetchItems.partition(bs.nonEmpty && _.getUrl.startsWith(bs.get.baseUrl))
```
rvanheest: use `map` and `getOrElse` here

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -367,8 +357,8 @@ object DepositHandler {
       }
   }
 
-  private def compareChecksumAgainstReferredBag(file: String, checksum: String, url: String)(implicit id: String, bs: BagStoreBase): Try[Unit] = {
-    val referredFile = getReferredFile(url, bs.baseUrl)
+  private def compareChecksumAgainstReferredBag(file: String, checksum: String, url: String)(implicit id: String, bs: Option[BagStoreSettings]): Try[Unit] = {
+    val referredFile = getReferredFile(url, bs.get.baseUrl)
```
rvanheest: How do you know that `bs` is not a `None` here? You don't! So don't use a `get`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -458,20 +448,20 @@ object DepositHandler {
...
       bag.getPayloadManifests
         .asScala
         .flatMap(_.asScala)
     })
 
-  private def getReferredBagDir(url: String)(implicit bs: BagStoreBase): File = {
+  private def getReferredBagDir(url: String)(implicit bs: Option[BagStoreSettings]): File = {
     //  http://deasy.dans.knaw.nl/aips/31aef203-55ed-4b1f-81f6-b9f67f324c87.2/data/x -> 31/aef20355ed4b1f81f6b9f67f324c87/2
-    val Array(uuid, version) = url.stripPrefix(bs.baseUrl)
+    val Array(uuid, version) = url.stripPrefix(bs.get.baseUrl)
```
rvanheest: same here

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -458,20 +448,20 @@ object DepositHandler {
...
       .split("/data").head.replaceAll("-", "")
       .split("\\.")
     val (topDir, uuidDir) = uuid.splitAt(3)
 
-    getFile(bs.baseDir, topDir, uuidDir, version)
+    getFile(bs.get.baseDir, topDir, uuidDir, version)
```
rvanheest: and here

Maybe you can give the `BagStoreBase` here as an implicit, rather than its `Option`. (**btw, it is a good thing to always try to avoid the `Option`, `Try`, etc. as an input parameter**). You can get it from somewhere, I guess

rvanheest (changes requested):
Quick look through it. Main issue is the use of `Option` here. Also have a look at [workshop 3](https://github.com/DANS-KNAW/course-scala/blob/master/src/main/scala/workshop3/03%20option.md) for pointers.

src/test/scala/nl/knaw/dans/api/sword2/ResolveFetchItemsSpec.scala
```diff
@@ -117,41 +117,27 @@ class ResolveFetchItemsSpec extends Sword2Fixture with BagStoreFixture {
...
-    implicit val bagStoreBase = BagStoreBase("non/existent/dir", baseUrl)
+    implicit val bagStoreSettings = Some(BagStoreSettings("non/existent/dir", baseUrl))
     copyToTargetBagDir(SIMPLE_SEQUENCE_A)
     val bagStoreCheck =  DepositHandler.checkBagStoreBaseDir()
     (the [IOException] thrownBy bagStoreCheck.get).getMessage should include("Bag store base directory")
   }
 
   it should "result in a Success when both bag-store base-dir and base-uri are not given, and there are no fetch.txt references to the bag-store"  in {
-    implicit val bagStoreBase = BagStoreBase("", "")
+    implicit val bagStoreSettings = None: Option[BagStoreSettings]
```
rvanheest: use `Option.empty[BagStoreSettings]` here and below

rvanheest (changes requested):
some more comments

src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
vesaakerman: What is the deeper problem?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: It is possible, but why is that needed?  Now we don't need to paas the settings object.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: I think that using `nonEmpty` is the most simple way and the most clear way to code this.  I looked into `getOrElse` and `map` and got constructions I was not happy with.  But perhaps I don't know how to do it neatly with `map`...

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: How would you use `map` or `getOrElse` here? I think that it just gets unnecessarily complicated.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: You should never come to this part of code if `bs` is `None`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: You should never come to this part of code if `bs` is `None`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: I tried it, when creating the implicit val `bagStoreSettings`, using getOrElse.  But then the type of the variable is `Product with Serializable`.  And also then you need to check then contents of `bagStoreSettings`.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: You're right; I left this `Option` in in my refactoring (https://github.com/vesaakerman/easy-sword2/pull/4).

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: I think `checkBagStoreBaseDir` only is supposed to be called whenever the bagstore is actually configured. So you can leave out the `Option` part completely and move that to `finalizeDeposit` (see https://github.com/vesaakerman/easy-sword2/pull/4/files#diff-ad17b49d50f1b93808fb326e8f79d0d8R87).

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: Well, you only need this partition if and only if the bagstore is actually configured. If not, you can just pass on the complete list. Saves you an extra iteration through the list. See https://github.com/vesaakerman/easy-sword2/pull/4/files#diff-ad17b49d50f1b93808fb326e8f79d0d8R262 for my refactoring.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: Not convinced by your statement. You can still get there with a `None` if you look carefully!
I refactored the code such that you give the `BagStoreSettings` rather than its `Option` here. Makes the code cleaner here! Given that you do not expect this to be happening, I created a `Failure` in https://github.com/vesaakerman/easy-sword2/pull/4/files#diff-ad17b49d50f1b93808fb326e8f79d0d8R252.

rvanheest:
See my PR (https://github.com/vesaakerman/easy-sword2/pull/4) for some refactorings regarding the `Option[BagStoreSettings]`.

src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
janvanmansum: The confusion between "state" and "statement".

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -81,28 +80,26 @@ object DepositHandler {
...
     log.info(s"[$id] Finalizing deposit")
-    // TODO pass on the combination object of baseDir and baseURL
-    implicit val baseDir: File = new File(settings.bagStoreBaseDir)
-    implicit val baseUrl: URI  = new URI(settings.bagStoreBaseUri)
+    implicit val optBagStoreSettings: Option[BagStoreSettings] = settings.bagStoreSettings
     val tempDir = new File(settings.tempDir, id)
 
     val result = for {
-      _        <- checkBagStoreBaseDir()
+      _        <- optBagStoreSettings.map(checkBagStoreBaseDir).getOrElse(Success(()))
```
janvanmansum: I still think that it is unnecessary and wasteful to check this for every request. Please, move this to the service initializer, as discussed earlier.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -81,28 +80,26 @@ object DepositHandler {
 
   def finalizeDeposit(mimeType: String)(implicit settings: Settings, id: String): Try[Unit] = {
     log.info(s"[$id] Finalizing deposit")
-    // TODO pass on the combination object of baseDir and baseURL
-    implicit val baseDir: File = new File(settings.bagStoreBaseDir)
-    implicit val baseUrl: URI  = new URI(settings.bagStoreBaseUri)
+    implicit val optBagStoreSettings: Option[BagStoreSettings] = settings.bagStoreSettings
```
janvanmansum: Is this implicit picked up by `checkBagVirtualValidity` even though the parameter in that function has a different name?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -253,7 +248,8 @@ object DepositHandler {
 
             if (missingFilesNotInFetchText.isEmpty)
               noFetchItemsAlreadyInBag(bagDir, itemsFromBagStore)
-                .flatMap(_ => validateChecksumsFetchItems(bag, itemsFromBagStore))
+                .flatMap(_ => bagStoreSettings.map(implicit bs => validateChecksumsFetchItems(bag, itemsFromBagStore))
```
janvanmansum, line 243: Another place where the implicit is defined with a different name than the parameter of the function that uses the implicit. @rvanheest : are there best practices for this? I find it confusing and I think it obfuscates the code.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -253,7 +248,8 @@ object DepositHandler {
...
+                  .getOrElse(Failure(new NoSuchElementException("BagStore is not configured - SOMETHING WENT WRONG, YOU SHOULD NEVER REACH THIS PART OF CODE!"))))
```
janvanmansum, line 244: Why could we never get here?

src/test/scala/nl/knaw/dans/api/sword2/ResolveFetchItemsSpec.scala
```diff
@@ -94,24 +99,43 @@ class ResolveFetchItemsSpec extends Sword2Fixture with BagStoreFixture {
...
+  }
+
+  it should "result in a Success when both bag-store base-dir and base-uri are not given, and there are no fetch.txt references to the bag-store"  in {
+    implicit val bagStoreSettings = Option.empty[BagStoreSettings]
+    copyToTargetBagDir(SIMPLE_SEQUENCE_A)
+    DepositHandler.checkFetchItemUrls(targetBagDir, urlPattern) shouldBe a[Success[_]]
+    DepositHandler.checkBagVirtualValidity(targetBagDir) shouldBe a[Success[_]]
+  }
+
+  it should "result in a Failure when both bag-store base-dir and base-uri are not given, and there are fetch.txt references to the bag-store"  in {
```
janvanmansum: I guess this is the test we talked about, that tries to download files from URL's that are no dereferenceable, that take a long time to fail, right?

janvanmansum (changes requested):
There are still some things I don't understand or that I don't agree with. See my comments. Please, explain or modify accordingly.

src/main/scala/nl/knaw/dans/api/sword2/DepositProperties.scala
janvanmansum: @vesaakerman:
Just to record the refactoring we just discussed in person:
- Rename the 'Statement' case class to 'DepositProperties' and add fields for all the properties in the `deposit.properties` file.
- Create and return objects of this class with the companion object's `apply` 
- Add a `getState`-method to the class that returns an enumeration value

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: I think that only the type of the implicit is checked? Correct @rvanheest ?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: This is something different from what we discussed earlier.  Here we have both `base-dir `and `base-url` present, but the path to the `base-dir` doesn't exist. Do you really want to have this checked in `ServiceInitializer`, which means that the whole service doesn't start when the directory doesn't exist?  What if when the service is started, the `base-dir` is there, but later for some reason it gets deleted?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: yup, implicits operate on types. So, you can only have one `Option[BagStoreSettings]` in scope. When this is the case, the value corresponding to that type is provided. Names are only used for/by the people that write them

rvanheest: btw. I forgot to remove the type of this value. It's not necessary here. It's up to you to leave it in or through it out.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: Because:
    val (**fetchItemsInBagStore**, itemsToResolve) = fetchItems.partition(bagStoreSettings.nonEmpty && _.getUrl.startsWith(bagStoreSettings.get.baseUrl))

If **fetchItemsInBagStore** is empty, we never come to this case:
 `case (itemsFromBagStore, false)`

src/test/scala/nl/knaw/dans/api/sword2/ResolveFetchItemsSpec.scala
vesaakerman: you are right, I'll comment it out.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
rvanheest: but you can't prove this on compile level. it would take a large refactoring to do this in another way. so we decided to do it this way. Afterward we decided to add a giant flag to the error message (for logging purposes), meaning that if we see this, we have definitely made a mistake somewhere...

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
janvanmansum: - I do want the service to refuse to start if preconditions for normal functioning are not satisfied.
- If for some reason the system gets messed up *while* it is running, there will of course be an error when trying to access the non-existent directory. The question is: do we have to check above preconditions before *every* request? That seems a bit much for something that is very unlikely to happen.

vesaakerman: Now the existence of the directory is checked once for each deposit.  If that is too much, then we have to 'build' something to make it happen less frequently.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -81,28 +80,26 @@ object DepositHandler {
...
     log.info(s"[$id] Finalizing deposit")
-    // TODO pass on the combination object of baseDir and baseURL
-    implicit val baseDir: File = new File(settings.bagStoreBaseDir)
-    implicit val baseUrl: URI  = new URI(settings.bagStoreBaseUri)
+    implicit val bagStoreSettings = settings.bagStoreSettings
     val tempDir = new File(settings.tempDir, id)
 
     val result = for {
-      _        <- checkBagStoreBaseDir()
+      _        <- bagStoreSettings.map(checkBagStoreBaseDir).getOrElse(Success(()))
```
janvanmansum: @vesaakerman : please remove, as discussed in live code review.

src/main/scala/nl/knaw/dans/api/sword2/servlets/ServiceInitializer.scala
```diff
@@ -86,13 +85,22 @@ class ServiceInitializer extends ServletContextListener {
       case _ => throw new RuntimeException(s"Invalid authentication settings: ${config.getString("auth.mode")}")
     }
     val urlPattern = Pattern.compile(config.getString("url-pattern"))
-    // TODO: make use of bag-store optional through configuration
     val bagStoreBaseUri = config.getString("bag-store.base-url") // TODO: make File, check existence
     val bagStoreBaseDir = config.getString("bag-store.base-dir") // TODO: make File, check existence
+    var bagStoreSettings = Option.empty[BagStoreSettings]
```
janvanmansum, line 91: @vesaakerman : please, use isBlank from apache commons, to avoid possible NPEs.

src/test/scala/nl/knaw/dans/api/sword2/ResolveFetchItemsSpec.scala
```diff
@@ -94,24 +99,43 @@ class ResolveFetchItemsSpec extends Sword2Fixture with BagStoreFixture {
...
+    (the [IOException] thrownBy bagStoreCheck.get).getMessage should include("Bag store base directory")
   }
+
+  it should "result in a Success when both bag-store base-dir and base-uri are not given, and there are no fetch.txt references to the bag-store"  in {
+    implicit val bagStoreSettings = Option.empty[BagStoreSettings]
+    copyToTargetBagDir(SIMPLE_SEQUENCE_A)
+    DepositHandler.checkFetchItemUrls(targetBagDir, urlPattern) shouldBe a[Success[_]]
+    DepositHandler.checkBagVirtualValidity(targetBagDir) shouldBe a[Success[_]]
+  }
+//   COMMENTED OUT BECAUSE IT TOOK TOO LONG TO GET RESPONSE
```
janvanmansum: @vesaakerman : please, remove.

src/test/scala/nl/knaw/dans/api/sword2/ResolveFetchItemsSpec.scala
```diff
@@ -58,31 +58,36 @@ class ResolveFetchItemsSpec extends Sword2Fixture with BagStoreFixture {
   it should "result in a Failure when a required file is missing"  in {
     copyToTargetBagDir(REQUIRED_FILE_MISSING)
     DepositHandler.checkFetchItemUrls(targetBagDir, urlPattern) shouldBe a[Success[_]]
-    DepositHandler.checkBagVirtualValidity(targetBagDir) shouldBe a[Failure[_]]
+    val validity = DepositHandler.checkBagVirtualValidity(targetBagDir)
```
janvanmansum, line 62: @vesaakerman : better use http://doc.scalatest.org/1.9.1/index.html#org.scalatest.Inside to directly check the contents of the `Failure` object: 
```scala
 inside(validity) {
      case Failure(e) =>
        e shouldBe a[InvalidDepositException]
        e.getMessage should include("Bag does not have bagit.txt")
    }
```

janvanmansum (changes requested):
See comments.

PR #74, opened 2016-11-30
EASY-1126 extend SWORD statement with the item-uri's of archived files

vesaakerman:
fixes EASY-1126

#### When applied
* it will store the `bag-id` and `filepaths` of the deposited files in `deposit.properties` when the bag is deposited
* after the bag has been archived, it will add the` bag-id` and `filepaths` into the `AtomStatement` when the client enquires the deposit state  

* deposit a bag (/send-simple.sh my/bag.zip http://deasy.dans.knaw.nl/sword2/collection/1)
* ask state (./get.sh  http://deasy.dans.knaw.nl/sword2/statement/mendeltest-...)

easy-                      | [PR#](PRlink)     | 

@DANS-KNAW/easy


src/main/scala/nl/knaw/dans/api/sword2/StatementManagerImpl.scala
```diff
@@ -38,8 +39,18 @@ class StatementManagerImpl extends StatementManager {
...
+          properties.resources.foreach(resources => setResources(statement, resources))
         statement
       case Failure(t) => throw new SwordError(404)
     }
   }
+
+  private def setResources(statement: AtomStatement, resources: PropertiesResources): Unit =  {
+    val bagIdResource = new ResourcePart(resources.bagId.trim)
+    bagIdResource.setMediaType("multipart/related;type=bag")
+    val filePathsResources = resources.filePaths.map(resource => { val resourcePart = new ResourcePart(resource.trim); resourcePart.setMediaType("multipart/related;type=file"); resourcePart})
```
rvanheest: not entirely sure about this lambda... I don't think you need the `resourcePart`

janvanmansum: At any rate, don' put everything on one line.

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
```diff
@@ -453,4 +453,31 @@ object DepositHandler {
 
     getFile(bagStoreSettings.baseDir, topDir, uuidDir, version)
   }
+
+  def writeToProperties(bagDir: File)(implicit settings: Settings, id: String): Try[Unit] = Try {
+    for {
+      bag <- getBag(bagDir)
+      resources <- getResources(bag)
```
janvanmansum, line 460: This was not the intention. `easy-ingest-flow` should set the resources when they are actually created in archival storage, not `easy-sword2`. At this point we cannot be sure that the archival copy has actually been created.

src/main/scala/nl/knaw/dans/api/sword2/package.scala
```diff
@@ -43,6 +43,8 @@ package object sword2 {
 
   case class BagStoreSettings(baseDir: String, baseUrl: String)
 
+  case class PropertiesResources(bagId: String, filePaths: List[String])
```
janvanmansum: Instead of bagId and filePaths it should be bagUri and fileUris. When sending an update, the client will include URIs in the `fetch.txt`.

janvanmansum (changes requested):
Not quite right yet. Please, see comments.

src/main/scala/nl/knaw/dans/api/sword2/package.scala
vesaakerman: Do you mean just renaming the parameters, or also giving another type (URI)?
About the filePaths/fileUri's: I had understood that we don't give the complete file Uri's, but a path relative to the bagUri.

janvanmansum: Both.

About the other question: the [src](https://tools.ietf.org/html/rfc4287#section-4.1.3.2) attribute must contain an IRI, so in our case a URL.

vesaakerman: I looked into changing the type of the parameters from String to URL, but I am wondering if we should do it.  These values are input for org.swordapp.server.ResourcePart, where they are given as String (even though representing an IRI).  We store these values in Properties, also as Strings, and thus have to convert them when reading from or writing to Properties file. To what avail actually?  Wouldn’t it be more logical just to use the same type as in ResourcePart or in Properties, namely String?

src/main/scala/nl/knaw/dans/api/sword2/DepositHandler.scala
vesaakerman: As agreed in our discussion, I left this code in DepositHandler.  There is a check in StatementManagerImpl that takes care that the client gets to see the file uris only when the state is ARCHIVED

PR #81, opened 2017-03-06
NO JIRA ISSUE: Edit-IRI returns 404 for invalid, but existing deposit

janvanmansum:
The Edit-IRI returned a 404 when it was invalid, because the code only looked in the inbox
to see if it could find the deposit. When a deposit is invalid, however, it is never moved
to the inbox, but instead left in the temp directory. Now the client uses the DepositProperties
object, which looks in both places.


janvanmansum:
@DANS-KNAW/easy for review.

PR #82, opened 2017-03-28
NO JIRA ISSUE: pospone deleting zips until after validation

janvanmansum:
The zip files received from the client where deleted as soon as the combined
zip was unpacked. The disadvantage of this was that is was then impossible to
examine the zip file if the bag was found to be invalid. Examining the zip file
could be useful to trace back the problem to the client (or determine that the
the error was caused on the server).

I have now moved the step to delete the zip files to after the validation. It will
only be done if the bag is found to be VALID.

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -117,10 +117,11 @@ object DepositHandler {
 
   private def removeZipFiles(bagDir: File): Try[Unit] = Try {
     log.debug("Removing zip files")
-    bagDir.listFiles().toList.filter(f => isPartOfDeposit(f) && f.isFile).foreach {
-      f =>
-        log.debug(s"Removing $f")
-        deleteQuietly(f)
+    for (file <- bagDir.listFiles().toList
+         if isPartOfDeposit(file)
+         if file.isFile) {
+      log.debug(s"Removing $file")
+      deleteQuietly(file)
     }
   }
 
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -83,16 +83,17 @@ object DepositHandler {
   def finalizeDeposit(mimeType: String)(implicit settings: Settings, id: String): Try[Unit] = {
     log.info(s"[$id] Finalizing deposit")
     implicit val bagStoreSettings = settings.bagStoreSettings
-    val tempDir = new File(settings.tempDir, id)
+    val depositDir = new File(settings.tempDir, id)
 
     val result = for {
       _        <- extractBag(mimeType)
-      bagDir   <- getBagDir(tempDir)
+      bagDir   <- getBagDir(depositDir)
       _        <- checkFetchItemUrls(bagDir, settings.urlPattern)
       _        <- checkBagVirtualValidity(bagDir)
       props    <- DepositProperties(id)
       _        <- props.setState(SUBMITTED, "Deposit is valid and ready for post-submission processing")
       _        <- props.save()
+      _        <- removeZipFiles(depositDir)
       dataDir  <- moveBagToStorage()
     } yield ()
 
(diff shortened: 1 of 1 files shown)


janvanmansum:
@DANS-KNAW/easy for review

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -114,6 +115,15 @@ object DepositHandler {
     }
   }
 
+  private def removeZipFiles(bagDir: File): Try[Unit] = Try {
+    log.debug("Removing zip files")
+    bagDir.listFiles().toList.filter(f => isPartOfDeposit(f) && f.isFile).foreach {
```
rvanheest: maybe you could make this better readable if you put it in a for-loop:

```scala
for (file <- bagDir.listFiles().toList
     if isPartOfDeposit(file)
     if file.isFile) {
    log.debug(s"Removing $file")
    deleteQuietly(file)
}
```

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -161,9 +168,8 @@ object DepositHandler {
   }
 
   private def getBagDir(depositDir: File): Try[File] = Try {
-    val depositFiles = depositDir.listFiles.filter(isPartOfDeposit)
-    if (depositFiles.length != 1) throw InvalidDepositException(depositDir.getName, s"A deposit package must contain exactly one top-level file, which must be a directory, number found: ${depositFiles.size}")
-    if (depositFiles(0).isFile) throw InvalidDepositException(depositDir.getName, s"Deposit package contained one top-level file, but it was not a directory")
+    val depositFiles = depositDir.listFiles.filter(_.isDirectory)
```
rvanheest, line 172: why don't you check the `isPartOfDeposit` part anymore?

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -161,9 +168,8 @@ object DepositHandler {
   }
 
   private def getBagDir(depositDir: File): Try[File] = Try {
-    val depositFiles = depositDir.listFiles.filter(isPartOfDeposit)
-    if (depositFiles.length != 1) throw InvalidDepositException(depositDir.getName, s"A deposit package must contain exactly one top-level file, which must be a directory, number found: ${depositFiles.size}")
-    if (depositFiles(0).isFile) throw InvalidDepositException(depositDir.getName, s"Deposit package contained one top-level file, but it was not a directory")
```
rvanheest, line 166: I guess removing this line is because you now do a `_.isDirectory` check now?

rvanheest (commented):
**General question:** moving the _delete zip files_ step to after the validation means that the bag is not deleted if the bag is not valid. That's the intention of the PR indeed. However, who is going to delete it then? Does this mean that we require vast amounts of extra storage to accommodate for users sending in creppy stuff? Who is going to go through all this stuff? I guess you need to do this manually. Perhaps I'm incorrect, but I see lots of extra labor heading our way...

PaulBoon (commented):
Maybe move the bags that failed to a 'mortuary' folder for 'post-mortem' inspection?

rvanheest:
@PaulBoon that's actually not a bad idea! It is indeed better to move them there than to keep them hanging around in the same directory as all other deposits that come in.
@janvanmansum would that be difficult to add?

PaulBoon (commented):
B.T.W. you can always link this 'mortuary' to /dev/null

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
janvanmansum: Right, we now just ignore any regular files, because it is no longer prohibited for them to be in the deposit.

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
janvanmansum: `isPartOfDeposit` was defined as "is not the `deposit.properties` file". However, we now just require that there be only one directory in the deposit, namely the unzipped bag. Any other files are ignore, including `deposit.properties` and any zip files, because we want to be able to keep those around longer.

janvanmansum:
Yes, as discussed during the dev team, that would be a useful feature to add once we introduce a database for the deposit state metadata (for lack of a better).

rvanheest (approved):

PR #83, opened 2017-04-15
EASY-1254

janvanmansum:
Added locale config to service start-up scripts, so that files are read and
written with UTF-8 encoded file names

fixes EASY-1254

### DIFF
--- a/src/main/assembly/dist/bin/easy-sword2-service-start
+++ b/src/main/assembly/dist/bin/easy-sword2-service-start
@@ -3,6 +3,8 @@
 BINPATH=`command readlink -f $0 2> /dev/null || command grealpath $0 2> /dev/null`
 APPHOME=`dirname \`dirname $BINPATH \``
 
+# Set LC_ALL to a locale with UTF-8 to make sure non-ASCII file names are written correctly to the file system (see: EASY-1254).
+LC_ALL=en_US.UTF-8 \
 java -Dlogback.configurationFile=$APPHOME/cfg/logback-service.xml \
      -Dapp.home=$APPHOME \
-     -cp $APPHOME/bin/easy-bag-store.jar nl.knaw.dans.easy.sword2.Sword2Service
+     -cp $APPHOME/bin/easy-sword2.jar nl.knaw.dans.easy.sword2.Sword2Service
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -17,6 +17,7 @@ package nl.knaw.dans.easy.sword2
 
 import java.io.{File, IOException}
 import java.net.{MalformedURLException, URL, UnknownHostException}
+import java.nio.charset.StandardCharsets
 import java.nio.file._
 import java.nio.file.attribute.{BasicFileAttributes, PosixFilePermissions}
 import java.util.regex.Pattern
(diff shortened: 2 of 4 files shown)


janvanmansum:
@DANS-KNAW/easy for review.

src/main/assembly/dist/bin/easy-sword2-service-start
```diff
@@ -3,6 +3,8 @@
 BINPATH=`command readlink -f $0 2> /dev/null || command grealpath $0 2> /dev/null`
 APPHOME=`dirname \`dirname $BINPATH \``
 
+# Set LC_ALL to a locale with UTF-8 to make sure non-ASCII file names are written correctly to the file system (see: EASY-1254).
+LC_ALL=en_US.UTF-8 \
 java -Dlogback.configurationFile=$APPHOME/cfg/logback-service.xml \
      -Dapp.home=$APPHOME \
-     -cp $APPHOME/bin/easy-bag-store.jar nl.knaw.dans.easy.sword2.Sword2Service
+     -cp $APPHOME/bin/easy-sword2.jar nl.knaw.dans.easy.sword2.Sword2Service
```
rvanheest, line 10: Wauw! That is an interesting catch! This script apparently never gets executed...

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -126,7 +126,11 @@ object DepositHandler {
   }
 
   private def extractBag(mimeType: String)(implicit settings: Settings, id: String): Try[File] = {
-    def extract(file: File, outputPath: String): Unit = new ZipFile(file.getPath).extractAll(outputPath)
+    def extract(file: File, outputPath: String): Unit = {
+      val zip = new ZipFile(file.getPath)
+      zip.setFileNameCharset("UTF-8")
```
rvanheest: I would more prefer using `StandardCharsets.UTF_8.name()` than just a hard-coded String.

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -126,7 +126,11 @@ object DepositHandler {
   }
 
   private def extractBag(mimeType: String)(implicit settings: Settings, id: String): Try[File] = {
-    def extract(file: File, outputPath: String): Unit = new ZipFile(file.getPath).extractAll(outputPath)
+    def extract(file: File, outputPath: String): Unit = {
+      val zip = new ZipFile(file.getPath)
```
rvanheest: General tip on API's that require you to call extra setters after the constructor call: expand the constructor as shown below. I'm not saying you should use this in this piece of code; it's more a general remark. I've been using this trick when I need to call a constructor, call some setters and return the object afterwards.

```
new ZipFile(file.getPath) {
    setFileNameCharset(StandardCharsets.UTF_8.name())
}
```

rvanheest (approved):
Looks good!

src/main/assembly/dist/bin/easy-sword2-service-start
janvanmansum: It will once we upgrade to RHEL 7, and start using `systemd`.

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
janvanmansum: So, you are creating an anonymous sub-class of `ZipFile`, which set this to UTF-8? Looks better indeed, also nice to use `StandardCharsets`!

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
janvanmansum: 👍

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
rvanheest: You could say that indeed. It's kind of extending the constructor.

PR #84, opened 2017-05-09
Disk space

janvanmansum:
NO JIRA ISSUE.

Quick fix. This adds checks to see if there is enough disk space to go ahead before concatenating partial deposits and before unzipping the deposit. If there isn't the deposit is rejected, so that smaller deposits can continue to be processed.

Warning: the code isn't pretty, but as I work on this code I realize it needs to be thoroughly cleaned up anyway.

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -15,36 +15,40 @@
  */
 package nl.knaw.dans.easy.sword2
 
-import java.io.{File, IOException}
-import java.net.{MalformedURLException, URL, UnknownHostException}
+import java.io.{ File, IOException }
+import java.net.{ MalformedURLException, URL, UnknownHostException }
 import java.nio.charset.StandardCharsets
 import java.nio.file._
-import java.nio.file.attribute.{BasicFileAttributes, PosixFilePermissions}
+import java.nio.file.attribute.{ BasicFileAttributes, PosixFilePermissions }
+import java.util
 import java.util.regex.Pattern
-import java.util.{Collections, NoSuchElementException}
+import java.util.{ Collections, NoSuchElementException }
 
 import gov.loc.repository.bagit.FetchTxt.FilenameSizeUrl
 import gov.loc.repository.bagit.transformer.impl.TagManifestCompleter
 import gov.loc.repository.bagit.utilities.SimpleResult
 import gov.loc.repository.bagit.verify.CompleteVerifier
 import gov.loc.repository.bagit.writer.impl.FileSystemWriter
-import gov.loc.repository.bagit.{Bag, BagFactory, FetchTxt}
+import gov.loc.repository.bagit.{ Bag, BagFactory, FetchTxt }
 import net.lingala.zip4j.core.ZipFile
+import net.lingala.zip4j.model.FileHeader
 import nl.knaw.dans.easy.sword2.State._
-import nl.knaw.dans.lib.error.{CompositeException, TraversableTryExtensions}
+import nl.knaw.dans.lib.error.{ CompositeException, TraversableTryExtensions }
 import org.apache.abdera.i18n.iri.IRI
 import org.apache.commons.codec.digest.DigestUtils
 import org.apache.commons.io.FileUtils._
-import org.joda.time.{DateTime, DateTimeZone}
-import org.slf4j.{Logger, LoggerFactory}
-import org.swordapp.server.{Deposit, DepositReceipt, SwordError}
+import org.joda.time.{ DateTime, DateTimeZone }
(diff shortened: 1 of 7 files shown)


janvanmansum:
@DANS-KNAW/easy for review.

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -106,6 +110,14 @@ object DepositHandler {
           _ <- props.setState(INVALID, msg)
           _ <- props.save()
         } yield ()
+      case RejectedDepositException(_, msg, cause) =>
+        log.error(s"[$id] Rejected deposit", cause)
+        for {
+          _ <- removeZipFiles(depositDir) // Currently, the only reason for SWORD 2 to reject a deposit, is insufficient disk space, so we clean up, here.
+          props <- DepositProperties(id)
```
rvanheest, line 117: I see code duplication with the case above. Only difference is the `INVALID`/`REJECTED` state. Can we make a method out of this?

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -127,6 +139,31 @@ object DepositHandler {
...
+        Failure(RejectedDepositException(id, "Not enough disk space to process deposit.",
+          new IllegalStateException(s"Required disk space for unzipping: ${uncompressedSize + settings.marginDiskSpace} (including ${settings.marginDiskSpace} margin). Available: $availableDiskSize")))
+      else Success(())
+    }
+
+    def checkDiskspaceForMerging(files: Seq[File]): Try[Unit] = {
+      val requiredSpace = files.map(_.length).sum
+      files.headOption.map {
+        f =>
+          val availableDiskSize = Files.getFileStore(f.toPath).getUsableSpace
```
rvanheest, line 158: is this duplicated from the function above? I guess this makes up for another function of type `Path => Try[Unit]`?

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -106,6 +110,14 @@ object DepositHandler {
           _ <- props.setState(INVALID, msg)
           _ <- props.save()
         } yield ()
+      case RejectedDepositException(_, msg, cause) =>
+        log.error(s"[$id] Rejected deposit", cause)
+        for {
+          _ <- removeZipFiles(depositDir) // Currently, the only reason for SWORD 2 to reject a deposit, is insufficient disk space, so we clean up, here.
```
rvanheest, line 116: If this is the only reason, why don't you call it an `InsufficientDiskSpaceException`? Makes the intent more clear.

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -159,13 +196,18 @@ object DepositHandler {
           files.foreach(file => {
             if (!file.isFile)
               throw InvalidDepositException(id, s"Inconsistent dataset: non-file object found: ${file.getName}")
+            checkAvailableDiskspace(file).get
```
rvanheest, line 199: I'll do a separate PR for this whole `Try` block, because this looks so bad!

Something that is out of reach for putting a comment: in line 219 you're using a `listFiles`, while you have defined a `listFilesSafe` to make sure null-pointers get handled properly. Shouldn't you use the latter there as well? Also look for other uses of `listFiles` in this project while you're at it.

rvanheest (changes requested):
I can only agree that the code is a mess in some places. This needs careful work before it becomes an easy-app that no one wants to touch!

Here are some comments to start you off. I'll do a separate PR/amend for some of the code.

rvanheest:
See https://github.com/janvanmansum/easy-sword2/pull/2

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
janvanmansum: I want to keep the exceptions aligned with the external states, and and the same time limit those external states to a relatively small and well-defined number. Both `FAILED` and `REJECTED` make sense, here. `FAILED` because the problem lies on the server, and `REJECTED` because we actually reject the deposit for being too large to process. Also, we  decided earlier to only show a generic message to the client in case of `FAILED` deposits, but I think that in this case it makes more sense to actually tell the client what the problem is, so that they may refrain from sending more large (but in fact valid) deposits.

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
janvanmansum: Sure, something to put on the list for a Great Clean-Up 👍

PR #93, opened 2017-08-18
EASY-1291

rvanheest:
Fixes EASY-1291

#### When applied it will
* update dependent role version
* fix the yml file
* put a better reasonable default in the properties
* fixed the logback file
* fixed the rpm scripts

@DANS-KNAW/easy for review

### DIFF
--- a/src/main/ansible/vagrant.yml
+++ b/src/main/ansible/vagrant.yml
@@ -32,16 +32,6 @@
 - hosts: "test"
   become: yes
   tasks:
-  - name: Create directories
-    file:
-      path: "{{ item }}"
-      recurse: yes
-      state: directory
-      mode: "0777"
-    with_items:
-        - /var/opt/dans.knaw.nl/tmp/easy-sword2
-        - /var/opt/dans.knaw.nl/spool/sword2-deposits
-
   - name: Install package
     yum:
       name: dans.knaw.nl-easy-sword2
--- a/src/main/ansible/vagrant.yml
+++ b/src/main/ansible/vagrant.yml
@@ -23,29 +23,31 @@
     easy_test_users_email: "{{ deployer_email }}"
     easy_test_users_ldapadmin_password: "{{ easy_ldap_dir_ldapadmin_password }}"
   roles:
-    - dans.local-test-vm-base
-    - dans.easy-ldap-dir
     - dans.capture-deployer-email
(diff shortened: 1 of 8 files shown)


src/main/assembly/dist/cfg/application.properties
```diff
@@ -4,7 +4,7 @@ deposits.permissions=rwxrwx---
 tempdir=/var/opt/dans.knaw.nl/tmp/easy-sword2
 # Require at least some 2 G  spare disk space
 tempdir.margin-available-diskspace-mb=2000
-base-url=http://yourdomain-here.org/
+base-url=http://test.dans.knaw.nl/
```
janvanmansum: I disagree. When creating "reasonable defaults" picture the scenario of someone installing the rpm via a `yum install`, but you don't know anything else about them or their environment. The base-url of their server is therefore something that you really cannot predict.

src/main/ansible/vagrant.yml
```diff
@@ -23,29 +23,31 @@
...
-      yum:
-        name: dans.knaw.nl-easy-sword2
-        state: latest
+  - name: Create directories
+    file:
+      path: "{{ item }}"
+      recurse: yes
+      state: directory
+      mode: "0777"
+    with_items:
```
janvanmansum: I am in favor of keeping the creation of these default directories in the RPM post install.

janvanmansum (changes requested):
I want to keep the creation of default dirs in the RPM. The rest seems OK.

src/main/assembly/dist/cfg/application.properties
rvanheest: mmm, maybe you're right. I was thinking the 'reasonable default' should cater towards the local vm. Also I don't expect anyone else to use this software, other than us...

src/main/ansible/vagrant.yml
rvanheest: I'm not. I think both configuration of these directories and creating them should be left to the deployer, e.g. in ansible/vagrant. Imagine the deployer wishes to put these directories in a different place, now he first has to remove the old directory (which he might not have the rights to?), then create the new directory and set the configurations correctly. I think we should keep the rpm install for what it currently does: stopping the service, installing the software and restarting the service.

janvanmansum (approved):

PR #96, opened 2017-12-20
EASY-1414: creationTimestamp

gulcinermis:
Fixes EASY-1414

#### When applied it will
* add creation.timestamp to deposit.properties

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
@@ -90,7 +90,7 @@ class DepositProperties(depositId: String, depositorId: Option[String] = None)(i
       .getOrElse(Failure(new IllegalStateException("Deposit without state")))
   }
 
-   /**
+  /**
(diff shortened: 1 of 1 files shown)


src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
```diff
@@ -119,6 +121,8 @@ class DepositProperties(depositId: String, depositorId: Option[String] = None)(i
   def getLastModifiedTimestamp: Option[FileTime] = {
     modified
   }
+
+
```
jo-pol: Why these extra new lines? 
Also please import https://github.com/DANS-KNAW/easy-module-archetype/blob/master/dans-intellij-codestyles.xml in your intellij _preferences - editor - code-style_ and apply the code formatting

janvanmansum: 👍 , but we checked this and it turns out our style does not remove 2 extra lines before a brace. Maybe we should change the style.

rvanheest (approved):

PR #102, opened 2018-07-05
NO JIRA Configurable niceness

janvanmansum:
#### When applied it will
*  Make the [niceness](https://linux.die.net/man/1/nice) with which the service runs configurable through the initd script.

In response to an issue on the production server about `easy-sword2` gobbling up too much processor time.

@DANS-KNAW/easy

### DIFF
--- a/src/main/assembly/dist/install/easy-sword2-initd.sh
+++ b/src/main/assembly/dist/install/easy-sword2-initd.sh
@@ -9,7 +9,8 @@
 #                    and should be placed in /etc/init.d
 
 NAME="easy-sword2"
-EXEC="/usr/bin/jsvc"
+NICENESS=0
+EXEC="nice -n $NICENESS /usr/bin/jsvc"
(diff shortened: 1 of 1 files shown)


rvanheest:
Can you go into some more detail of what this is, what it does and why we need it?

janvanmansum:
@rvanheest : What is does: see the link I provided. It was needed because of `easy-sword2` getting too much of the processor on production.

rvanheest:
I saw the link, but that was just general. What I wanted to know was why **we** use it. That's clearer now.

Is this also required on the rest of the servers! So not only production but also deasy, teasy and aeasy?

janvanmansum:
@rvanheest It is not really required on the other servers, but I don't thing it will hurt. In my understanding it will only make a difference if there is a lot of competition for CPU time.

lindareijnhoudt (approved):

PR #103, opened 2018-09-03
EASY-1695:  Make sure only depositors can check status of deposit

lindareijnhoudt:
Fixes EASY-1695


janvanmansum (approved):
One remark: the client now gets a 404 when it is not authenticated. The could be on purpose, as it is nobody's business that this deposit exists, except the depositor's, but it could also be because the existing code takes some questionable shortcuts with respect to error handling. Let's leave it as is for this release.

PR #105, opened 2018-11-06
EASY-1779: Add identifier.doi.registered=no to DepositProperties

stephanHogenboom:
- [x] do manual test to verify that the field is added

Fixes EASY-1779-1781

the DepositProperties class

create a deposit and check if its properties file is created with the according field


rvanheest (approved):

lindareijnhoudt:
are you still planning on adding the unit test?

stephanHogenboom:
Changed it to a manual test, too much production code needed to be altered to unit test this change properly

PR #106, opened 2018-11-23
Easy-1804 set state of a deposit to uploaded, once all files are uploaded to the sword2 directory

stephanHogenboom:
Fixes EASY-Easy 1804 uploaded
#### When applied it will
* will give states the status uploaded instead of finalizing once the file is uploaded

easy-                      | [PR#7_deposit-lib](https://github.com/DANS-KNAW/dans-deposit-lib/pull/7)     | ..
easy-                      | [EMD_PR#36](https://github.com/DANS-KNAW/easy-manage-deposit/pull/36)     | ..

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -248,7 +248,7 @@ object DepositHandler {
       log.info(s"[$id] Scheduling deposit to be finalized")
       val result = for {
         props <- DepositProperties(id)
-        _ <- props.setState(FINALIZING, "Deposit is being reassembled and validated")
+        _ <- props.setState(UPLOADED, "Deposit upload has been completed.")
         _ <- props.save()
       } yield ()
       result.get // Trigger exception if properties could not be updated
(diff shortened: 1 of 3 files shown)


src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -469,13 +469,10 @@ object DepositHandler {
   }
 
   def isOnPosixFileSystem(file: File): Boolean = {
-    try {
+    Try {
```
rvanheest: good catch! That's better indeed.
**challenge**: you could even make this into a one-liner if you want. hint: use [Try.fold](https://www.scala-lang.org/api/current/scala/util/Try.html#fold[U](fa:Throwable=%3EU,fb:T=%3EU):U)

rvanheest (approved):

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
stephanHogenboom: `Try(Files.getPosixFilePermissions(file.toPath)).fold(_ => false, _ => true)?`

rvanheest: you rock! 👍

stephanHogenboom: Thanks for the tip!

PR #109, opened 2018-11-26
EASY-1802 give write permission to group for deposits 

stephanHogenboom:
Fixes EASY-1802

#### When applied it will
* give write acces to the group
#### How should this be manually tested?

try to modify/ delete a deposit and its content on  deasy/ teasy in the sword tmp folder

PR #110, opened 2018-11-27
EASY-1805: reschedule deposits on not enough disk space

rvanheest:
Fixes EASY-1805

#### When applied it will
* use type aliases for `DepositId` and `MimeType`
* let `depositProcessingStream` take a tuple of `DepositId` and `MimeType` rather than `DepositId` and `Deposit`
* rename `RejectedDepositException` to `NotEnoughDiskSpaceException`
* when `NotEnoughDiskSpaceException` is encountered, reschedule the deposit, rather than `REJECT` it.

@DANS-KNAW/easy for review

* [x] wait for #109 to be merged

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/Authentication.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/Authentication.scala
@@ -78,7 +78,7 @@ object Authentication {
   }
 
   @throws(classOf[SwordAuthException])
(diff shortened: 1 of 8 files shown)


rvanheest:
@lindareijnhoudt this PR is finished. Please keep in mind that with merging this PR you should also change the flow diagram accordingly. There isn't a `REJECTED` state in `easy-sword2` anymore!

PR #116, opened 2018-12-17
EASY-1864 add file permissions to new deposits

stephanHogenboom:
- [x] manual check

Fixes EASY-1864

#### When applied it will
* give the right permissions to new deposits so operators can edit and delete the data if needed

easy- sword2                     | [PR#109](https://github.com/DANS-KNAW/easy-sword2/pull/109)     | ..


src/main/scala/nl.knaw.dans.easy.sword2/FilesPermission.scala
```diff
@@ -43,8 +44,9 @@ object FilesPermission {
         case cce: ClassCastException => log.error("No file permission elements in set", cce)
         case ioe: IOException => log.error(s"Could not set file permissions on $path", ioe)
         case se: SecurityException => log.error(s"Not enough privileges to set file permissions on $path", se)
+        case iae: IllegalArgumentException => log.error(s"Incorrect permissions input string: $permissions, could not set permission on $path", iae)
         case e => log.error(s"Unexpected exception on $path", e)
-      } getOrElse FileVisitResult.TERMINATE
+      } unsafeGetOrThrow
```
jo-pol, line 54: Though not part of this issue, I'd rather see something with Vistor in the class name, it doesn't do anything recursively. @rvanheest an opinion about something like `FilePermissionVistor`?

stephanHogenboom: im also pretty sure i cannot just throw here

rvanheest: @jo-pol what about `RecursiveFilePermissionVisitor`?

stephanHogenboom: i can live with that name

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -105,18 +105,19 @@ object DepositHandler {
 
   def handleDeposit(deposit: Deposit)(implicit settings: Settings, id: DepositId): Try[DepositReceipt] = {
     val contentLength = deposit.getContentLength
+    println(s"deposit file name is ${deposit.getFilename}")
```
rvanheest:
```suggestion
```

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -110,16 +110,31 @@ object DepositHandler {
...
+      _ <- if (contentLength > -1) assertTempDirHasEnoughDiskspaceMarginForFile(contentLength)
+           else Success(())
+      _ <- copyPayloadToFile(deposit, payload)
+      _ <- doesHashMatch(payload, deposit.getMd5)(id)
       _ <- handleDepositAsync(deposit)
-      _ <- FilesPermission.changePermissionsRecursively(deposit.getFile, settings.depositPermissions, id) // set file permissions after continued deposit is finished
       dr = createDepositReceipt(settings, id)
       _ = dr.setVerboseDescription("received successfully: " + deposit.getFilename + "; MD5: " + deposit.getMd5)
     } yield dr
+    receipt
```
rvanheest: I think you can inline this return 😃

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -110,16 +110,31 @@ object DepositHandler {
...
 
     val payload = Paths.get(settings.tempDir.toString, id, deposit.getFilename.split("/").last).toFile
-    for {
+    val depositDir = Paths.get(settings.tempDir.toString, id).toFile
+    val receipt: Try[DepositReceipt] = extractPayloadAndGetDepositReceipt(deposit, contentLength, payload, depositDir)
+    setFilePermissions(settings, id, depositDir) // ensure all files in deposit are accessible for the deposit group
+    receipt
+  }
+
+  private def setFilePermissions(settings: Settings, id: DepositId, depositDir: File) = {
```
rvanheest: It would really help us in the future if you add return types for private methods as well. I know it is not necessary for the Scala compiler, but for us 'humans' it is quite nice to see what the return type is.

(same for the method below)

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -110,16 +110,31 @@ object DepositHandler {
     }
 
     val payload = Paths.get(settings.tempDir.toString, id, deposit.getFilename.split("/").last).toFile
-    for {
+    val depositDir = Paths.get(settings.tempDir.toString, id).toFile
+    val receipt: Try[DepositReceipt] = extractPayloadAndGetDepositReceipt(deposit, contentLength, payload, depositDir)
+    setFilePermissions(settings, id, depositDir) // ensure all files in deposit are accessible for the deposit group
```
rvanheest: looks like you're throwing away a potential `Failure` here... use a for-comprehension in combination with the line above.

stephanHogenboom: Yes, i was looking a that as well. I want that setFilesPermissions is always executed, whether extractPayloadAndGetDepositReceipt is a success or not

rvanheest: that's interesting... because `Try` is failfast, as you know.
However, why do you want `setFilesPermissions` to be executed always? Is there any file to set the permissions for when `extractPayloadAndGetDepositReceipt` fails?

stephanHogenboom: It is a bit messy, during manual testing i found out it can leave a zip file, a bag with an invalid md5 or both.

rvanheest: It's not very clean, but you could do something like this:

```scala
val x: Try[Int] = f() // f :: () -> Try[Int]
val y: Try[Unit] = g() // g :: () -> Try[Unit]

for {
  a <- x
  _ <- y
} yield a
```

by running `f` and `g` first, you perform their respective side effects first and capture the result (`Success`/`Failure`) in a value. Afterwards, you merge the 2 results using a for-comprehension. Not a clean way, but at least it's something...

In this case, document carefully why you do it this way (perhaps also refer to this discussion), because otherwise someone (probably me 😉) is going to refactor this!

rvanheest: What you could also do is:

```scala
// f :: () -> Try[Int]
// g :: () -> Try[Unit]

f()
  .recoverWith { case e => g().flatMap(_ => Failure(e)) }
  .flatMap(i => g().map(_ => i))
```

this is probably a better solution!

rvanheest: Or:

```scala
// f :: () -> Try[Int]
// g :: () -> Try[Unit]

f() match {
  case Success(i) => g().map(_ => i)
  case Failure(e) => g().flatMap(_ => Failure(e))
}
```

stephanHogenboom: think the last one is the most elegant

rvanheest: thought so too. Looks great the way you did it!

rvanheest (approved):

PR #121, opened 2019-01-29
EASY-1876 easy-solr4files-index faalt op filenames met sommige vreemde tekens

vesaakerman:
Fixes EASY-1876

#### When applied it will
* take care of a few reserved characters in file names, that would cause an error at some point of ingest. Filenames are checked for reserved characters and an error message is shown if found. 

See the attached document:
[special characters.pdf](https://github.com/DANS-KNAW/easy-sword2/files/2807569/special.characters.pdf)

Example of a bag with not-reserved special characters:
[not-reserved-special-characters-bag.zip](https://github.com/DANS-KNAW/easy-sword2/files/2807877/not-reserved-special-characters-bag.zip)

Example of a bag with reserved special characters:
[reserved-special-characters-bag.zip](https://github.com/DANS-KNAW/easy-sword2/files/2807883/reserved-special-characters-bag.zip)

On your local machine go to easy-sword2-dans-examples directory. To test a bag containing reserved characters, give a command:

`./run.sh Simple http://deasy.dans.knaw.nl/sword2/collection/1 user001 user001 src/main/resources/examples/reserved-special-characters-bag`

easy-                      | [PR#](PRlink)     | ..


janvanmansum (commented):
The issue description was about fixing `easy-bag-index` not `easy-sword2`, so why are you doing a PR for `easy-sword2`. Also, I was not able to reproduce the issue, partly because there was no "steps to reproduce", but the example that I created myself did not produce an error. Also, I don't understand how it could make `easy-bag-index` fail, as this service does not access the payload files as far as I know. @vesaakerman: did you reproduce the problem?

janvanmansum (commented):
OK, so I finally understood that this is not about `easy-bag-index` but about `easy-solr4files-index`. Still not about `easy-sword2` though. It is good to write down your proposed solution, but then please shared it before you start programming.

vesaakerman:
Surely I have reproduced the problem.  And found many more, so as I write in the attached document.

janvanmansum:
I still think the problem needs to be solved in `easy-solr4files-index`. Note that a file item in the bag-store has a percent-encoded path. It looks like `easy-solr4files-index` does not apply the percent-encoding in its request. See also: https://dans-knaw.github.io/easy-bag-store/03_definitions.html#item-id.

vesaakerman:
The method UrlEscapers.urlPathSegmentEscaper() is used... But I'll look into this (after my 2 days off:))

vesaakerman:
Should we remove ';' from safe characters in URL_PATH_SEGMENT_ESCAPER?

janvanmansum:
I am closing this issue, as this is not the correct project to solve it in.

PR #122, opened 2019-02-01
NO-JIRA: Sword2 logback-service parser error

stephanHogenboom:
#### When applied it will
* now the sword2 logs a parser error instead of the class name (Richard Kruithof found this fix wil resolve this).

### DIFF
--- a/src/main/assembly/dist/cfg/logback-service.xml
+++ b/src/main/assembly/dist/cfg/logback-service.xml
@@ -8,7 +8,7 @@
             <maxHistory>30</maxHistory>
         </rollingPolicy>
         <encoder>
-            <pattern>[%date{ISO8601}] %-5level [..%t8.8] %msg%n%ex</pattern>
+            <pattern>[%date{ISO8601}] %-5level [%8.8t] %msg%n%ex</pattern>
         </encoder>
     </appender>
     <root level="warn">


rvanheest (commented):
Why don't we do `%t` instead of this `8.8` formatting business. I wanna see the whole thread name...

stephanHogenboom:
Would be fine by me, shouldn't we discuss that with @janvanmansum  and @lindareijnhoudt ?

rvanheest:
I'm not even sure why the thread name was added in the first place (see https://github.com/DANS-KNAW/easy-sword2/commit/9ab21e615f3d619dcb261bc254f67fbae02fec73#diff-b4af71bdc07bb66a118a57639463f577), but we should definitely discuss this...

janvanmansum:
@stephanHogenboom @rvanheest @lindareijnhoudt : the thread name was added to enable us to see in what thread things are happening. The deposit parts are received by Jetty-threads and the the finalization is done on one background thread. Unfortunately the thread name is rather long, hence the idea to truncate it. You really only need to be able to tell the apart.

I don't remember exactly in what context we decided we needed the thread name, so I am fine with removing it altogether and see if we run into a situation where we feel we need it.

janvanmansum:
I am hearing no other comments. I'll close this one and remove the thread name for now.

PR #123, opened 2019-02-04
NO JIRA. Refactoring: rename contentType to easy-sword2.client-message.content-type

janvanmansum:
NO JIRA.

Rename a temporary property  in `deposit.properties` and clean it up after use, to avoid confusion.

### Some background
The property is used to record what type of messages the sword client has been sending, i.e. `application/zip` or `application/octet-stream`. This information is needed in the "finalizing" stage. Since deposits can be pushed back onto the finalization queue if there is not enough room for them to be processed, this means that the information must be saved to `deposit.properties`, so that a restart of `easy-sword2` will not cause those deposits never to be processed.

- [x] Test

#### When applied it will
* Rename `contentType` to `easy-sword2.client-message.content-type`
* Make sure the `easy-sword2` still understands the old property, too.
* Ensure that `easy-sword2` cleans up after successfully finalizing, as the property will not be useful after that point.

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
@@ -99,8 +99,8 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
   }
 
   def removeClientMessageContentType(): Try[DepositProperties] = Try {
-    properties.setProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY, null)
-    properties.setProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY_OLD, null) // Also clean up old contentType property if still found
+    properties.clearProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY)
+    properties.clearProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY_OLD) // Also clean up old contentType property if still found
     this
   }
 
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
@@ -21,6 +21,7 @@ import java.nio.file.attribute.FileTime
 import nl.knaw.dans.easy.sword2.State.State
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.configuration.PropertiesConfiguration
+import org.apache.commons.lang.StringUtils
 import org.joda.time.{ DateTime, DateTimeZone }
 
 import scala.util.{ Failure, Success, Try }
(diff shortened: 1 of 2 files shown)


src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
```diff
@@ -92,15 +93,24 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
...
-    properties.setProperty("contentType", contentType)
+  def setClientMessageContentType(contentType: String): Try[DepositProperties] = Try {
+    properties.setProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY, contentType)
     this
   }
 
-  def getContentType: Try[String] = {
-    Option(properties.getString("contentType"))
+  def removeClientMessageContentType(): Try[DepositProperties] = Try {
+    properties.setProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY, null)
```
rvanheest: I think you mean to use `clearProperty` here and below?

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
```diff
@@ -92,15 +93,24 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
...
+    properties.setProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY_OLD, null) // Also clean up old contentType property if still found
+    this
+  }
+
+  def getClientMessageContentType: Try[String] = {
+    Seq(properties.getString(CLIENT_MESSAGE_CONTENT_TYPE_KEY),
+      properties.getString(CLIENT_MESSAGE_CONTENT_TYPE_KEY_OLD)) // Also look for old contentType to support pre-upgrade deposits
+      .find(StringUtils.isNotBlank)
```
rvanheest, line 110: given that `properties.getString` returns `null` if the key is not present, doesn't this fail?

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
```diff
@@ -139,6 +149,8 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
 
 object DepositProperties {
   val FILENAME = "deposit.properties"
+  val CLIENT_MESSAGE_CONTENT_TYPE_KEY_OLD = "contentType" // for backwards compatibility
+  val CLIENT_MESSAGE_CONTENT_TYPE_KEY = "easy-sword2.client-message.content-type"
```
rvanheest, line 153: So, should we add this property to [easy-specs](https://github.com/DANS-KNAW/easy-specs/blob/master/deposit-directory/deposit-directory.md#depositproperties)?

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
janvanmansum: Good one! I had been looking for a `deleteProperty` or `removeProperty` ...

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
janvanmansum: Do you mean like this?
```scala
import org.apache.commons.lang.StringUtils
Seq(null, "test").find(StringUtils.isNotBlank)
```
The produces `Some(test)` in a Scratch book

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
janvanmansum: 👍

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
rvanheest: 👍 great! Than `isNotBlank` takes `null` into account perfectly. Thanks for checking. That solves my concern.

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
janvanmansum: See: https://github.com/DANS-KNAW/easy-specs/pull/64/files

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
```diff
@@ -92,15 +93,24 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
...
+  def removeClientMessageContentType(): Try[DepositProperties] = Try {
+    properties.clearProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY)
+    properties.clearProperty(CLIENT_MESSAGE_CONTENT_TYPE_KEY_OLD) // Also clean up old contentType property if still found
+    this
+  }
+
+  def getClientMessageContentType: Try[String] = {
+    Seq(properties.getString(CLIENT_MESSAGE_CONTENT_TYPE_KEY),
+      properties.getString(CLIENT_MESSAGE_CONTENT_TYPE_KEY_OLD)) // Also look for old contentType to support pre-upgrade deposits
+      .find(StringUtils.isNotBlank)
```
rvanheest, line 110:
```diff
- .find(StringUtils.isNotBlank)
- .map(Success(_))
+ .collectFirst { s if StringUtils.isNotBlank(s) => Success(s) }
```

janvanmansum: Sorry, I don't find that more readable.

PR #126, opened 2019-03-07
EASY-1929 return stack traceless error to client

stephanHogenboom:
Fixes EASY-1929

#### When applied it will
* Easy-sword2 will no longer return a stack trace in the error message to the client

- [x] ~~change error messages?~~
- [x] ~~change 405 to a 404 with other error message?~~


src/main/scala/nl.knaw.dans.easy.sword2/CollectionDepositManagerImpl.scala
```diff
@@ -21,7 +21,7 @@ import java.nio.file.Paths
 import nl.knaw.dans.easy.sword2.DepositHandler._
 import nl.knaw.dans.easy.sword2.State._
 import org.apache.commons.lang.StringUtils._
-import org.swordapp.server._
+import org.swordapp.server.{ SwordError, _ }
```
rvanheest:
```suggestion
import org.swordapp.server._
```

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -574,7 +574,7 @@ object DepositHandler {
 
   def doesHashMatch(zipFile: File, MD5: String)(implicit id: DepositId): Try[Unit] = {
     log.debug(s"[$id] Checking Content-MD5 (Received: $MD5)")
-    lazy val fail = Failure(new SwordError("http://purl.org/net/sword/error/ErrorChecksumMismatch"))
+    lazy val fail = Failure(new SwordError("http://purl.org/net/sword/error/ErrorChecksumMismatch", 400))
```
rvanheest: shouldn't this be left at the (default) `412`? That's what the issue says, and also what `ErrorChecksumMismatch` means (see [`UriRegistry`](https://github.com/DANS-KNAW/easy-sword2-lib/blob/cb9abc3cb519632166637b3bfea90dc1ef237af9/src/main/java/org/swordapp/server/UriRegistry.java#L54) and [`ErrorDocument`](https://github.com/DANS-KNAW/easy-sword2-lib/blob/cb9abc3cb519632166637b3bfea90dc1ef237af9/src/main/java/org/swordapp/server/ErrorDocument.java#L43)).

In fact, it better be `new SwordError(UriRegistry.ERROR_CHECKSUM_MISMATCH)` instead of this `String`.

src/main/scala/nl.knaw.dans.easy.sword2/SwordConfig.scala
```diff
@@ -19,7 +19,7 @@ import org.swordapp.server.SwordConfigurationDefault
 
 class SwordConfig extends SwordConfigurationDefault {
   override def getAuthType: String = "Basic"
-
+  override def returnStackTraceInError(): Boolean = false //TODO would it be desirable to do this with the application.properties file?
```
rvanheest: @janvanmansum @lindareijnhoudt what do you think?

On the one hand I don't think it's a bad idea to put it in `application.properties`: we may want to have these stacktraces on our development machines and not on the ACT/production machines.

On the other hand, I do value the consistency of output between the different servers, which is broken when we configure them differently.

rvanheest (changes requested):

rvanheest (approved):
Looks good to me!
If you find more of these URLs that can be replaced by preconfigured constants in this project, feel free to replace them!

rvanheest (approved):

PR #128, opened 2019-05-16
EASY-2082 Dryad ZipException

janvanmansum:
Fixes EASY-2082

- [x] Test on `deasy`
   - [x] With files that have non-ASCII file names

#### When applied it will
* Use better.files instead of zip4j to extract the deposit

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -289,9 +289,9 @@ object DepositHandler {
     }
 
     def extract(file: File, outputPath: String): Unit = {
-      val f = better.files.File(file.getAbsolutePath)
+      import better.files._
       implicit val charset: Charset = StandardCharsets.UTF_8
-      f unzipTo better.files.File(outputPath)
+      file.toScala unzipTo outputPath.toFile
     }
 
     def getSequenceNumber(f: File): Int = {
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -17,7 +17,7 @@ package nl.knaw.dans.easy.sword2
 
 import java.io.{ File, IOException }
 import java.net.{ MalformedURLException, URL, UnknownHostException }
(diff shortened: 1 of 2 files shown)


src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -289,9 +289,9 @@ object DepositHandler {
     }
 
     def extract(file: File, outputPath: String): Unit = {
-      new ZipFile(file.getPath) {
-        setFileNameCharset(StandardCharsets.UTF_8.name)
-      }.extractAll(outputPath)
+      val f = better.files.File(file.getAbsolutePath)
```
rvanheest: use an import statement above instead of `better.files.File`. Or even better

```scala
import better.files._

file.toScala // transforms Java IO File to better-files File
```

Same for the `String outputPath` below: use `outputPath.toFile`.

For reference, see https://github.com/pathikrit/better-files/blob/b4ad1a095eddf8bbec592812bd38bb4a335bb712/core/src/main/scala/better/files/Implicits.scala#L31-L48

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -289,9 +289,9 @@ object DepositHandler {
     }
 
     def extract(file: File, outputPath: String): Unit = {
-      new ZipFile(file.getPath) {
-        setFileNameCharset(StandardCharsets.UTF_8.name)
-      }.extractAll(outputPath)
+      import better.files._
```
rvanheest, line 292: you should know better!!! 😉
Imports go on top of the file...

rvanheest (approved):
looks good apart from the import

src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
janvanmansum: Well, in this case I think that is not such a good idea. It will cause a lot of changes because we would need to import `java.io.File` under an alias.

rvanheest: 👍never thought about that! Good point.

PR #131, opened 2019-06-27
add exception message to genericErrorMessage

rvanheest:
Add error message to `genericErrorMessage` to at least get a glimpse of what kind of error occurred.

@DANS-KNAW/easy for review

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -127,11 +127,11 @@ object DepositHandler extends BagValidationExtension {
     } yield dr
   }
 
-  def genericErrorMessage(implicit settings: Settings, id: DepositId): String = {
+  def genericErrorMessage(msg: String)(implicit settings: Settings, id: DepositId): String = {
     val mailaddress = settings.supportMailAddress
     val timestamp = DateTime.now(DateTimeZone.UTC).toString
 
-    s"""The server encountered an unexpected condition.
+    s"""The server encountered an unexpected condition: $msg.
        |Please contact the SWORD service administrator at $mailaddress.
        |The error occurred at $timestamp. Your 'DepositID' is $id.
     """.stripMargin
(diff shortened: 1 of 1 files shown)


janvanmansum (commented):
I am not sure that we should do this. Depending on what the error  message contains this may expose internal implementation details to the client, e.g. file system paths. We already log the full exception on line 207 in DepositHandler, so we can find out what went wrong from there (until the logs are removed, course, but we can revise the log clean-up policies if necessary).

rvanheest:
Perhaps you're right. My thinking was that the error message might be useful to the client as well, as they might come up with a solution for it on their own. @lindareijnhoudt any thoughts?

janvanmansum:
> Perhaps you're right. My thinking was that the error message might be useful to the client as well, as they might come up with a solution for it on their own. @lindareijnhoudt any thoughts?

The generic error occurs when the deposit fails, right? I would say that with FAILED deposits the client is not responsible for solving the problem.

rvanheest:
> I would say that with FAILED deposits the client is not responsible for solving the problem.

That's not really the case. `FAILED` occurs when no other exceptions match in the error handling.

https://github.com/DANS-KNAW/easy-sword2/blob/b653adfb85a9fb8d406cff3db05afe5663f3a935/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala#L206-L213

Hence I would not say that "_the client is not responsible for solving the problem_". It might very well be the case that the client caused this error by providing invalid input, which made our application 'crash'. `FAILED` is then still the correct value, but I would say we show this error to the user, such that he/she is informed on the kind of error. With that I hope to shine some light on what the user possibly did wrong.

rvanheest:
Not sure if anecdotal evidence counts, but yesterday I received a question from one of our new SWORD2 customers, asking about this. It was a typical user error (unzipping the bag), which was marked as a FAILED deposit due to the fact that it was caught in the `NonFatal(e)` case shown above. It would have been nice if we could have presented the customer with at least some information on what the error was.

janvanmansum:
I still think `FAILED` should be our responsibility. If it turns out the problem was due to invalid input, then we need to update the code, so that it correctly handles this situation by setting the deposit to `REJECTED`.

PR #137, opened 2019-10-16
EASY-2326. easy-sword2: unexpected 'error while setting filePermissions'

janvanmansum:
Fixes EASY-2326

#### When applied it will
* Make sure handing off the deposit to the finalizing thread is the last action that accesses the
  deposit

DepositHandler

This is hard to verfiy, as the problem only ocurred once. I have alread verified that it doesn't break anything too obvious, by doing one deposit via SWORD2 on `deasy`; this succeeded.

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
@@ -106,12 +106,7 @@ object DepositHandler extends BagValidationExtension {
     // to ensure all files in deposit are accessible for the deposit group, the method setFilePermissions is always
     // executed (regardless of whether extractAndValidatePayloadAndGetDepositReceipt was successful or not).
     // Otherwise operators don't have the proper permissions to clean or fix the invalid zip files or bags that might stay behind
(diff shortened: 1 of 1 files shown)


src/main/scala/nl.knaw.dans.easy.sword2/DepositHandler.scala
```diff
@@ -106,12 +106,7 @@ object DepositHandler extends BagValidationExtension {
     // to ensure all files in deposit are accessible for the deposit group, the method setFilePermissions is always
     // executed (regardless of whether extractAndValidatePayloadAndGetDepositReceipt was successful or not).
     // Otherwise operators don't have the proper permissions to clean or fix the invalid zip files or bags that might stay behind
-    extractAndValidatePayloadAndGetDepositReceipt(deposit, contentLength, payload, depositDir) match {
```
janvanmansum, line 109: @rvanheest : do you remember if postponing the change-permission-step until after handing off the deposit to the finalizing thread had a particular reason? This construction below seems a bit convoluted, anyway.

rvanheest (approved):

PR #142, opened 2020-03-13
EASY-1945: upgrade dependencies

pfrolke:
Fixes EASY-1945

#### When applied it will
* upgrade parent POM version
* remove self-contained dependency version tags
* move unique dependency versions that aren't included in a parent POM to properties
* change `lines.toList` to `linesIterator.toList` due to https://github.com/scala/bug/issues/11125
* remove ZipFile setFilenameCharset UTF8, since it is already the default value

repo                       | PR
-------------------------- | -----------------
easy-                      | [PR#](PRlink)

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/servlets/EasySword2Servlet.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/servlets/EasySword2Servlet.scala
@@ -26,6 +26,6 @@ class EasySword2Servlet(version: String) extends ScalatraServlet
 
   get("/") {
     contentType = "text/plain"
-    logResponse(Ok(s"EASY SWORD2 is running v$version."))
+    Ok(s"EASY SWORD2 is running v$version.")
   }
 }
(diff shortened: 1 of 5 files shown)


src/main/scala/nl.knaw.dans.easy.sword2/servlets/EasySword2Servlet.scala
```diff
@@ -26,6 +26,6 @@ class EasySword2Servlet(version: String) extends ScalatraServlet
 
   get("/") {
     contentType = "text/plain"
-    Ok(s"EASY SWORD2 is running v$version.").logResponse
+    logResponse(Ok(s"EASY SWORD2 is running v$version."))
```
rvanheest: I think the response is logged twice now (after the upgrade of `dans-scala-lib`). Could you test if removing this whole `logResponse` still logs the response?

```suggestion
    Ok(s"EASY SWORD2 is running v$version.")
```

rvanheest:
please fix the merge conflict

rvanheest (approved):

rvanheest:
Note to self: before releasing this module, bump the `easy-sword2-lib` dependency!

PR #150, opened 2020-06-29
EASY-2739: Add urn-nbn self-link to Sword Statement

lindareijnhoudt:
Fixes EASY-2739

#### When applied it will
* Add urn-nbn self-link to Sword Statement

Submit a deposit, retrieve the sword-statement. 
It has two self-links, one to the DOI, one to the URN:NBN

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
@@ -142,6 +142,11 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
       .map(_.toString)
   }
 
+  def getUrn: Option[String] = {
+    Option(properties.getProperty("identifier.urn"))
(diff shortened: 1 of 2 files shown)


vesaakerman (approved):

src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
```diff
@@ -142,6 +142,11 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
       .map(_.toString)
   }
 
+  def getUrn: Option[String] = {
```
jo-pol, line 145: Did you create an issue to add this to https://github.com/DANS-KNAW/easy-deposit-properties

lindareijnhoudt: good one. the URN:NBN is already in the deposit.properties, so on that front we are set

PR #156, opened 2021-08-11
DD-563: Implement rule 1.2.7 (Organizational-Id) of the Dans Bagit Profile in easy-sword2

vesaakerman:
Fixes DD-563

#### When applied it will
* write `dataverse.other-id` to properties if `Has-Organizational-Identifier` is given in bag-info.txt
* write `dataverse.other-id-version` to properties if `Has-Organizational-Identifier-Version` is given in bag-info.txt

easy-                      | [PR#](PRlink)     | ..

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
+++ b/src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
@@ -103,6 +103,11 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
     this
   }
 
+  def setOtherIdentifier(id: String, version: String): Try[DepositProperties] = Try {
+    if (id.nonEmpty) properties.setProperty("dataverse.other-id", id)
+    if (version.nonEmpty) properties.setProperty("dataverse.other-id-version", version)
+    this
+  }
 
   /**
    * Returns the state when the properties were loaded.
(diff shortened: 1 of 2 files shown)


src/main/scala/nl.knaw.dans.easy.sword2/DepositProperties.scala
```diff
@@ -103,6 +103,11 @@ class DepositProperties(depositId: DepositId, depositorId: Option[String] = None
     this
   }
 
+  def setOtherIdentifier(id: String, version: String): Try[DepositProperties] = Try {
+    if (id.nonEmpty) properties.setProperty("dataverse.other-id", id)
+    if (version.nonEmpty) properties.setProperty("dataverse.other-id-version", version)
```
jo-pol, line 108:
```suggestion
    id.foreach(properties.setProperty("dataverse.other-id", _))
    version.foreach(properties.setProperty("dataverse.other-id-version", _))
```

vesaakerman: `id` and `version` are Strings.  You could do `Option(id).foreach...`  but then you would get this when the strings are empty: 
```
dataverse.other-id =
dataverse.other-id-version =
```

When we leave the code as it is now,  these properties come into the properties file only when they are given in bag-info.txt

jo-pol: sorry, assumed it where options.

jo-pol (approved):

